PR #8, opened 2023-07-18
Also run image pushing on push trigger

DomAyre:
The change which introduces a check to skip the image pushing workflows if coming from an external fork also skips when the event is push, this fixes that


KenGordon (approved):
As discussed.

PR #20, opened 2023-08-01
dm-verity support for encfs

Jinnan-Guo:
Add dm-verity protection to the encfs


KenGordon:
Hello, could you introduce yourself? We know a bunch of people at Imperial so that would save some googling.

Please take a look at the writable fs PR (24 I think). It will probably be the case that yours has to follow on from that one.
You will need to do something to make the merkle tree etc. Are you doing that with a separate file or appending the tree after the FS - that is what we do for the image layers. Also some doc to describe the scheme. There are choices about doing the tree over the encrypted layer vs over the decrypted layer, it would be good to discuss those in a document.

cmd/remotefs/azurefs.go
```diff
@@ -324,7 +401,37 @@ func containerMountAzureFilesystem(tempDir string, index int, fs AzureFilesystem
 	}
 	logrus.Debugf("Device opened: %s", deviceName)
 
-	// 4) Mount block device as a read-only filesystem.
+	// 4) Config the decrypted filesystem in dm-verity format
```
KenGordon: Surely some check as to whether this is required or not should happen here?

Jinnan-Guo:
> Hello, could you introduce yourself? We know a bunch of people at Imperial so that would save some googling.

Thanks Ken. I am a PhD student from Imperial College, advised by Prof. Peter Pietzuch. We are working on an Azure confidential container-based project with Dr. Kapil Vaswani and Dr. Andrew Paverd. After discussion with Kapil and Andrew we decided to add dm-verity to encfs for integrity.

> 
> Please take a look at the writable fs PR (24 I think). It will probably be the case that yours has to follow on from that one. You will need to do something to make the merkle tree etc. Are you doing that with a separate file or appending the tree after the FS - that is what we do for the image layers. Also some doc to describe the scheme. There are choices about doing the tree over the encrypted layer vs over the decrypted layer, it would be good to discuss those in a document.

Yep I'll follow PR 24. We are considering to add dm-verity as an additional top-level option besides writeable encfs, since dm-verity does not support write. I'll add more details in document. Thanks.

cmd/remotefs/azurefs.go
Jinnan-Guo: Yes I think dm-verity should be an config option

PR #22, opened 2023-08-08
Update key release api version to 7.4

BryceDFisher:
Update the mHSM key release api version to 7.4 per mHSM direction.


hgarvison (approved):

stevendongatmsft (approved):

stevendongatmsft:
In order to import operators other than 'equals', the import key golang logic needs a change as well, as it currently bakes in the equals operator.

PR #25, opened 2023-08-11
DO NOT MERGE add retry when vcekcertchain is empty

stevendongatmsft:
This PR is a recreate of PR https://github.com/microsoft/archived-confidential-sidecar-containers/pull/45 

The title explains the PR. Original thought was to use exponential backoffs on retrying. No need to retry for more than one time because if retry for once does not work, something must be badly configured - then the failure ought to surface. We also do not want to risk Azure being blacklisted by AMD.

DO NOT MERGE because there was a change on the Attest function. So until I verify the issue of initialCerts not properly configured still persists, this PR should not be merged.


stevendongatmsft:
No longer needed

PR #27, opened 2023-08-14
Reduce docker layers

BryceDFisher:
Reduce the number of layers in the SKR and encfs container.  Also make the images smaller by disabling APK caching in both.

### DIFF
--- a/docker/skr/Dockerfile.skr
+++ b/docker/skr/Dockerfile.skr
@@ -1,11 +1,10 @@
 FROM alpine:3.17.1
 
-RUN apk update && apk add curl
+RUN apk update && apk add --no-cache curl
 
 COPY ./bin/skr ./bin/get-snp-report /bin/
 COPY skr.sh tests/*_client.sh tests/skr_test.sh /
-RUN mkdir -p /tests/skr; mv *_client.sh /tests/skr; mv skr_test.sh /tests/skr
-RUN chmod +x /*.sh /tests/skr/*.sh; date > /made-date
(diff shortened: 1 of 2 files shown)


hgarvison (approved):

docker/skr/Dockerfile.skr
```diff
@@ -1,11 +1,10 @@
 FROM alpine:3.17.1
 
-RUN apk update && apk add curl
+RUN apk update && apk add --no-cache curl
 
 COPY ./bin/skr ./bin/get-snp-report /bin/
 COPY skr.sh tests/*_client.sh tests/skr_test.sh /
-RUN mkdir -p /tests/skr; mv *_client.sh /tests/skr; mv skr_test.sh /tests/skr
-RUN chmod +x /*.sh /tests/skr/*.sh; date > /made-date
+RUN mkdir -p /tests/skr; mv *_client.sh /tests/skr; mv skr_test.sh /tests/skr && chmod +x /*.sh /tests/skr/*.sh
```
hgarvison, line 7: You should probably do this in the Dockerfile.debug too (if applicable)

BryceDFisher: I'm not worried about the total number of layers in the debug version.  This was to reduce the number of layers to work around a layer limit present in the kata-cc uvm.

stevendongatmsft (approved):

stevendongatmsft:
Reviewed and manually tested. Verified key can be released and file system was mounted.

PR #30, opened 2023-08-22
Logging updates

hgarvison:
Updates to add more logs, make logging uniform, and to have logfile & loglevel be set as environment variables in the ARM template instead of hardcoding the values.


cmd/azmount/filemanager/azure.go
```diff
@@ -70,71 +87,95 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
 
-	// Setup data downloader
+	// Setup data downloader and uploader
 	fm.downloadBlock = AzureDownloadBlock
+	fm.uploadBlock = AzureUploadBlock
+
+	return nil
+}
+
+func AzureUploadBlock(blockIndex int64, b []byte) (err error) {
```
KenGordon, line 151: Does this belong in this PR? It doesn't look like logging changes.

examples/skr/aci-arm-template.json
```diff
@@ -30,7 +30,16 @@
               "command": [
                 "/skr.sh"
               ],
-              "image": "mcr.microsoft.com/aci/skr:2.2",
```
KenGordon: This does not look right. I can't see what image this container is supposed to use now.

examples/skr/aci-arm-template.json
```diff
@@ -30,7 +30,16 @@
               "command": [
                 "/skr.sh"
               ],
-              "image": "mcr.microsoft.com/aci/skr:2.2",
+              "environmentVariables": [
+                {
+                  "name": "LogFile",
+                  "value": "/log.txt"
```
KenGordon: You can't do this as an environment variable without some, likely excessively hard, policy to make sure the log doesn't get splatted over some file we rely on.

A basic rule of thumb is to never expose anything that gives the outside control over internal UVM/container state.

cmd/attestation-container/attestation-container.go
```diff
@@ -20,12 +20,16 @@ import (
...
+
+	"github.com/sirupsen/logrus"
 )
 
 var (
 	socketAddress             = flag.String("socket-address", "/tmp/attestation-container.sock", "The socket address of Unix domain socket (UDS)")
 	platformCertificateServer = flag.String("platform-certificate-server", "", "Server to fetch platform certificate. If set, certificates contained in security context directory are ignored. Value is either 'Azure' or 'AMD'")
 	insecureVirtual           = flag.Bool("insecure-virtual", false, "If set, dummy attestation is returned (INSECURE: do not use in production)")
+	logLevel                  = flag.String("loglevel", "warning", "Logging Level: trace, debug, info, warning, error, fatal, panic.")
+	logFile                   = flag.String("logfile", "", "Logging Target: An optional file name/path. Omit for console output.")
```
KenGordon, line 30: Where does "console output" end up? Is that on the ACI log tab in the portal? If so then are you certain that nothing sensitive can ever be logged?

cmd/azmount/filemanager/azure.go
hgarvison: I branched from the R/W update so that I would have the changes I made there (I added some logs to debug with). I assumed the R/W PR would get merged before this.

examples/skr/aci-arm-template.json
BryceDFisher: If it is expressed as an environment variable, it would have to be captured in the security policy.  So long as the policy doesn't allow ".*", this would be safe against malicious changes.

cmd/attestation-container/attestation-container.go
BryceDFisher: We also have methods available via the security policy to outright block stdio from containers.  A customer that is worried about sensitive items in the log can enable this block.

BryceDFisher: That being said, we should strive to make sure that sensitive items are not logged via error or warn.  I would say that debug logs would be ok to have sensitive information so long as a customer has to opt into debug.

cmd/attestation-container/attestation-container.go
```diff
@@ -49,34 +53,41 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
...
 	reportFetcher := attest.NewAttestationReportFetcher()
 	reportBytes, err := reportFetcher.FetchAttestationReportByte(reportData)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to fetch attestation report: %s", err)
 	}
 
+	logrus.Info("Setting platform certificate...")
 	var platformCertificate []byte
 	if platformCertificateValue == nil {
+		logrus.Info("Deserializing attestation report...")
```
BryceDFisher: These are tracing style log items.  They don't add much context or help on their own.

cmd/attestation-container/attestation-container.go
```diff
@@ -86,66 +97,94 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
...
 	}
 }
 
 func main() {
 	flag.Parse()
+
+	// if logFile is not set, logrus defaults to stderr
+	if *logFile != "" {
+		// If the file doesn't exist, create it. If it exists, append to it.
+		file, err := os.OpenFile(*logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
```
BryceDFisher: We really wouldn't want the group or other to be able to write to the log.  `0644` for a log file at most.

cmd/attestation-container/attestation-container.go
```diff
@@ -86,66 +97,94 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
...
+		if err != nil {
+			logrus.Fatal(err)
+		}
+		defer file.Close()
+		multi := io.MultiWriter(file, os.Stderr)
```
BryceDFisher: I'm not sure I'm a fan of always enabling logging to stderr.  If a customer chooses to use a file, it should go there instead of stderr.  If you really want to allow multiple locations, then the customer should be able to explicitly list all output locations.

cmd/attestation-container/attestation-container.go
```diff
@@ -86,66 +97,94 @@ func (s *server) FetchAttestation(ctx context.Context, in *pb.FetchAttestationRe
...
 		}
 
+		logrus.Info("Getting UVM Information...")
 		uvmInfo, err := common.GetUvmInformation()
 		if err != nil {
-			log.Fatalf("Failed to get UVM information: %s", err)
+			logrus.Fatalf("Failed to get UVM information: %s", err)
 		}
 
+		logrus.Info("Setting platform certificate server...")
```
BryceDFisher: All of these are tracing messages...

cmd/azmount/filemanager/azure.go
```diff
@@ -30,19 +30,19 @@ func tokenRefresher(credential azblob.TokenCredential) (t time.Duration) {
...
+	logrus.Infof("Current token fields: %v", curentTokenFields)
 
 	payload, err := base64.RawURLEncoding.DecodeString(curentTokenFields[1])
 	if err != nil {
-		logrus.Errorf("Error decoding base64 token payload: %v", err)
+		logrus.Errorf("Error decoding base64 token payload: %s", err)
 		return 0
 	}
-	logrus.Infof("current token payload: %v", string(payload))
+	logrus.Infof("Current token payload: %s", string(payload))
```
BryceDFisher: Is there anything sensitive in this token?  Probably should move to debug if there's risk.

cmd/azmount/filemanager/azure.go
```diff
@@ -86,58 +87,59 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
+				logrus.Info("Can't obtain a token required for accessing private blobs. Will retry in case the ACI identity sidecar is not running yet...")
 				time.Sleep(3 * time.Second)
 				count++
 				if count == 20 {
-					return errors.Wrapf(err, "timeout of 60 seconds expired. could not obtained token")
+					return errors.Wrapf(err, "Timeout of 60 seconds expired. Could not obtain token")
 				}
 			} else {
-				logrus.Infof("token obtained: %s continuing", token.AccessToken)
+				logrus.Infof("Token obtained: %s \nContinuing...", token.AccessToken)
```
BryceDFisher: Logs shouldn't span multiple lines like this.  Also, this access token is the same as a username + password.  Tracing logging only for the contents.

cmd/azmount/filemanager/azure.go
```diff
@@ -86,58 +87,59 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
 				break
 			}
 		}
 
 		tokenCredential := azblob.NewTokenCredential(token.AccessToken, tokenRefresher)
-		logrus.Infof("token credential created: %s", tokenCredential.Token())
+		logrus.Infof("Token credential created: %s", tokenCredential.Token())
```
BryceDFisher: Tokens are the same as username + password.

cmd/azmount/filemanager/azure.go
```diff
@@ -86,58 +87,59 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
 		fm.blobURL = azblob.NewPageBlobURL(*u, azblob.NewPipeline(tokenCredential, azblob.PipelineOptions{}))
-		logrus.Infof("blob url created: %s", fm.blobURL)
+		logrus.Infof("Blob URL created: %s", fm.blobURL)
```
BryceDFisher: It looks like this URL is going to contain the access_token in it.  This should be a debug or tracing message.

cmd/azmount/filemanager/azure.go
BryceDFisher: Rebased this PR to point to the R/W branch instead of main.  This is now moot.

cmd/azmount/filemanager/cache_test.go
```diff
@@ -207,7 +207,7 @@ func Test_GetBlock_TestBounds(t *testing.T) {
 	ClearCache()
 
 	err, _ := GetBlock(-1)
-	if err.Error() != "invalid block index (-1)" {
+	if err.Error() != "Invalid block index (-1)" {
```
BryceDFisher, line 210: This requires refactoring of the `GetBlock` function, but it is much better to do error checking via a type than the contents of the error.  `BlockIndexError` and `if errors.Is(err, BlockIndexError)` is much more stable than the string checks.

cmd/azmount/filemanager/cache_test.go
```diff
@@ -219,7 +219,7 @@ func Test_GetBlock_TestBounds(t *testing.T) {
 		t.Errorf("GetBlock(255) should have succeeded: %s", err.Error())
 	}
 	err, _ = GetBlock(256)
-	if err.Error() != "block index over limit (256 > 255)" {
+	if err.Error() != "Block index over limit (256 > 255)" {
```
BryceDFisher, line 222: Also here

cmd/azmount/filemanager/cache_test.go
```diff
@@ -340,7 +340,7 @@ func Test_SetBlock_TestBounds(t *testing.T) {
 	data := GenerateRandomData(BLOCK_SIZE)
 
 	err := SetBlock(-1, 0, data)
-	if err.Error() != "invalid block index (-1)" {
+	if err.Error() != "Invalid block index (-1)" {
```
BryceDFisher, line 343: and here

cmd/azmount/main.go
```diff
@@ -53,93 +54,96 @@ func main() {
 
 	flag.Parse()
 
+	if *logFile != "" {
+		// If the file doesn't exist, create it. If it exists, append to it.
+		file, err := os.OpenFile(*logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0666)
```
BryceDFisher: `0644` at most for log files.

cmd/azmount/main.go
```diff
@@ -53,93 +54,96 @@ func main() {
...
+		if err != nil {
+			logrus.Fatal(err)
+		}
+		defer file.Close()
+		multi := io.MultiWriter(file, os.Stderr)
```
BryceDFisher: Same comment as before.

BryceDFisher: Also, be careful with multiwriter.  Each write happens serially in the order listed.  A slow file write will block all other writes.  An error to one output will cause the log to not be written to any later output.

pkg/attest/attest.go
```diff
@@ -21,50 +21,54 @@ type CertState struct {
 	Tcbm        uint64      `json:"tcbm"`
 }
 
+const (
+	SHA256LEN = 32
```
BryceDFisher: This doesn't need to be exported.  `sha256len` was ok.

pkg/attest/platform_cert_fetcher.go
```diff
@@ -166,33 +173,36 @@ func fetchWithRetry(requestURL string, baseSec int, maxRetries int) ([]byte, err
 		}
 		res, err := http.Get(requestURL)
 		if err != nil {
+			logrus.Debugf("fetch on retry %d: http.Get failed: %s", retryCount, err)
 			retryCount++
 			continue
 		}
 		if 200 <= res.StatusCode && res.StatusCode < 300 {
```
BryceDFisher, line 187: What's the point of the `&& < 300`?  The first check also enforces the second check.

pkg/attest/platform_cert_fetcher.go
```diff
@@ -166,33 +173,36 @@ func fetchWithRetry(requestURL string, baseSec int, maxRetries int) ([]byte, err
...
 		if err != nil {
+			logrus.Debugf("fetch on retry %d: http.Get failed: %s", retryCount, err)
 			retryCount++
 			continue
 		}
 		if 200 <= res.StatusCode && res.StatusCode < 300 {
 			// Got successful status code 2xx
 			defer res.Body.Close()
-			resBody, err := ioutil.ReadAll(res.Body)
+			resBody, err := io.ReadAll(res.Body)
```
BryceDFisher: Be careful with `io.ReadAll`.  A malicious proxy can overflow your memory by responding with unlimited bytes.  In general you should pass the reader back to the caller to deal with the response directly, or wrap it in a LimitReader: https://pkg.go.dev/io#LimitReader.

pkg/attest/platform_cert_fetcher.go
```diff
@@ -166,33 +173,36 @@ func fetchWithRetry(requestURL string, baseSec int, maxRetries int) ([]byte, err
...
 			retryCount++
 			continue
 		} else {
 			// Got status code that is not worth to retry
 			defer res.Body.Close()
-			resBody, err := ioutil.ReadAll(res.Body)
+			resBody, err := io.ReadAll(res.Body)
 			if err != nil {
-				return nil, fmt.Errorf("got error while handling non successful response with status code %d: %s", res.StatusCode, err)
+				return nil, errors.Errorf("got error while handling non successful response with status code %d: %s", res.StatusCode, err)
```
BryceDFisher, line 209: Error responses from servers will commonly include helpful information in the resonse body.  Even if there is an error in reading the response body, you may have already read some bytes that could be useful.

pkg/attest/platform_cert_fetcher.go
```diff
@@ -224,40 +237,48 @@ func (certFetcher CertFetcher) retrieveCertChain(chipID string, reportedTCB uint
 			vcekPEMBytes := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
 
 			// now retrieve the cert chain
+			logrus.Info("Fetching cert chain from AMD endpoint...")
 			uri = fmt.Sprintf(AmdCertChainRequestURITemplate, certFetcher.Endpoint, certFetcher.TEEType)
 			certChainPEMBytes, err := fetchWithRetry(uri, defaultRetryBaseSec, defaultRetryMaxRetries)
 			if err != nil {
-				return nil, reportedTCB, errors.Wrapf(err, "pulling AMD certchain response from get request failed")
+				return nil, reportedTCB, errors.Wrapf(err, "pulling AMD cert chain response from get request failed")
```
BryceDFisher: Context would be great, like from which URL did you pull?  This may already be captured in the function call.

pkg/common/http.go
```diff
@@ -64,7 +64,7 @@ func HTTPResponseBody(httpResponse *http.Response) ([]byte, error) {
 
 	// Pull out response body
 	defer httpResponse.Body.Close()
-	httpResponseBodyBytes, err := ioutil.ReadAll(httpResponse.Body)
+	httpResponseBodyBytes, err := io.ReadAll(httpResponse.Body)
```
BryceDFisher, line 67: It's a good idea to wrap this with a LimitReader first to prevent unlimited bytes being read.

pkg/common/token.go
```diff
@@ -54,7 +55,7 @@ func GetToken(ResourceId string, i Identity) (r TokenResponse, err error) {
 	// Unmarshall response body into struct
 	err = json.Unmarshal(httpResponseBodyBytes, &r)
 	if err != nil {
-		return r, errors.Wrapf(err, "unmarshalling authentication token response failed")
+		return r, errors.Wrapf(err, "unmarshalling authentication token response failed for %s", httpResponseBodyBytes)
```
BryceDFisher: This http response could include the returned token, which would be sensitive data.

pkg/skr/skr_test.go
```diff
@@ -294,7 +294,7 @@ func Test_AKV(t *testing.T) {
 
 			certstring: "MI",
 
-			expectedError: errors.New("decoding certificate from string failed"),
+			expectedError: errors.New("decoding x509 certificate from string failed: MI"),
```
BryceDFisher, line 297: It's much more stable to check based on an error type rather than the contents.

cmd/azmount/filemanager/azure.go
```diff
@@ -30,19 +30,19 @@ func tokenRefresher(credential azblob.TokenCredential) (t time.Duration) {
 	// JWT tokens comprise three fields. the second field is the payload (or claims).
 	// we care about the `aud` attribute of the payload
 	curentTokenFields := strings.Split(currentToken, ".")
```
SethHollandsworth: typo for `curentTokenFields` -> `currentTokenFields`

pkg/attest/platform_cert_fetcher.go
ksayid: I think the `&& < 300` ensures that we're only returning the response body of requests with status code 2xx (successful requests). Responses with other status codes are handled in the other conditionals. If we remove `&& < 300`, then a server error response (5xx), for example, wouldn't get a retry, even though it should.

cmd/azmount/filemanager/azure.go
ksayid: fixed, thanks!

pkg/attest/platform_cert_fetcher.go
BryceDFisher: The `<=` really through me off (laugh)

PR #32, opened 2023-08-23
Sanitise branch names for image tags

DomAyre:
Some branches, especially those created by dependabot, have characters which are illegal for image tags, so this PR replaces any illegal characters with "-"

### DIFF
--- a/.github/workflows/push_attestation_image.yml
+++ b/.github/workflows/push_attestation_image.yml
@@ -46,7 +46,9 @@ jobs:
           if [ ${{ github.event_name }} == "push" ]; then
             echo "branch_name=main" >> $GITHUB_OUTPUT
           else  
-            echo "branch_name=${{ github.head_ref }}" >> $GITHUB_OUTPUT
+            branch_name=${{ github.head_ref }}
+            branch_name=$(echo $branch_name | sed 's/[^a-zA-Z0-9]/-/g')
+            echo "branch_name=$branch_name" >> $GITHUB_OUTPUT
           fi
 
       - name: Build and Push Docker Image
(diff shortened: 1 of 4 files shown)


.github/workflows/push_attestation_image.yml
```diff
@@ -46,7 +46,9 @@ jobs:
           if [ ${{ github.event_name }} == "push" ]; then
             echo "branch_name=main" >> $GITHUB_OUTPUT
           else  
-            echo "branch_name=${{ github.head_ref }}" >> $GITHUB_OUTPUT
+            branch_name=${{ github.head_ref }}
+            branch_name=$(echo $branch_name | sed 's/[^a-zA-Z0-9]/-/g')
```
takuro-sato: From the code, it might be hard to see why this sed is necessary. So how about using a new variable image_name here?
`image_name=$(echo $branch_name | sed 's/[^a-zA-Z0-9]/-/g')`

.github/workflows/push_encfs_image.yml
```diff
@@ -51,6 +51,7 @@ jobs:
           else  
             branch_name=${{ github.head_ref }}
           fi
+          branch_name=$(echo $branch_name | sed 's/[^a-zA-Z0-9]/-/g')
```
takuro-sato: Do we need check for image name length just in case?

.github/workflows/push_attestation_image.yml
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/32/commits/67cd179d2c29f504acc713ce8af4656c9f5456c6

.github/workflows/push_encfs_image.yml
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/32/commits/22a887b0d36eab3af296f64f433dbf63637ccb25

takuro-sato (approved):

PR #34, opened 2023-08-24
Revert "Bump golang.org/x/crypto from 0.0.0-20201217014255-9d13527586…

DomAyre:
This reverts commit 6df8c74b045680a8def84602550a9a8cd565656b.

I prematurely merged the version bump PR once the encfs example had passed but the SKR example did fail. The version bump changes the format of the returned attestation report so this needs reverting and then potentially remerging with a fix to maintain the output format.


takuro-sato (approved):
Interesting that we are still using v.0.0.0 which are free to break.

PR #37, opened 2023-08-24
Output attestation reports as Hex

DomAyre:
Returns to previous behaviour


DomAyre:
I've tested this with the key_release example and it does return to the same format as the 2.3 release

KenGordon (approved):

PR #38, opened 2023-08-24
Feature/importkey octfile

SethHollandsworth:
Feature to align the symmetric key workflow with the asymmetric flow for the importkey process. Instead of passing in the symmetric key via a command line arg, it is now passed in as a file path.

This is important as the key will no longer be in the command history as it was prior. It also negates the need for the `hexstring.py` file. If this hexstring file should be deleted in the same PR, please let me know. I wasn't sure if it's being used somewhere else. Thanks!


tools/importkey/main.go
```diff
@@ -240,11 +240,12 @@ func main() {
...
 		var err error
 
-		if keyHexString == "" {
+		if keyHexFile == "" {
 			octKey = make([]byte, 32)
 			rand.Read(octKey)
 		} else {
-			octKey, err = hex.DecodeString(keyHexString)
+			// read string from file keyHexFile as hexstring
+			octKey, err = ioutil.ReadFile(keyHexFile)
```
hgarvison: ioutil.ReadFile is deprecated, switch to os.ReadFile

examples/encfs/generatefs/generatefs.sh
```diff
@@ -18,8 +18,6 @@ fi
 
 echo "Key in hex string format"
 
-python hexstring.py $keyFilePath
```
hgarvison, line 21: Go ahead and delete the hexstring.py file, since it's not used elsewhere

tools/importkey/main.go
SethHollandsworth: do you think it would be helpful to replace the other 2 spots in this file that use `ioutil.ReadFile` in this PR?

hgarvison: yes

hgarvison (approved):

PR #42, opened 2023-09-01
Get attestation report with kernel 6.x with Go

takuro-sato:
## Change
- [x] Add NewAttestationReportFetcher5 and NewAttestationReportFetcher6
- [x] Change behavior of NewAttestationReportFetcher so that it returns either resut of NewAttestationReportFetcher5  or NewAttestationReportFetcher6 when possible. Otherwise return error.
- [x] Use the new feature in the sidecars

## Test
- [x] Test attestation-container and related package in Confidential ACI with kernel 5.x
- [x] Test attest package  with 6.x
- [ ] Test attestation-container  with 6.x
  - [ ] My 6.x environment can't access the Internet (So can't use azure cache servers now)
  - Also it doesn't have local platform certs
 - [ ] Test the other sidecars with 5.x
 - [ ] Test the other sidecars with 6.x


takuro-sato:
Recreated this as https://github.com/microsoft/confidential-sidecar-containers/pull/43

PR #43, opened 2023-09-01
Get attestation report with kernel 6.x with Go

takuro-sato:
## Change
- [x] Add NewAttestationReportFetcher5 and NewAttestationReportFetcher6
- [x] Change behavior of NewAttestationReportFetcher so that it returns either resut of NewAttestationReportFetcher5  or NewAttestationReportFetcher6 when possible. Otherwise return error.
- [x] Use the new feature in the sidecars

## Test
- [x] Test attestation-container and related package in Confidential ACI with kernel 5.x (using docker/attestation-container/deploy_and_test.sh)
- [x] Test attest package  with 6.x
- [ ] Test attestation-container  with 6.x
  - [ ] My 6.x environment can't access the Internet (So can't use azure cache servers now)
  - Also it doesn't have local platform certs
 - [x] Test the other sidecars with 5.x (https://github.com/microsoft/confidential-aci-examples/actions/runs/6073266351)
 - [ ] Test the other sidecars with 6.x
   - [x] Check POST /raw/attest is working


KenGordon (approved):
Looks ok to me. Note that we cannot test this with a 6.1 kernel quite yet but this needs to be merged to make that easier. This is a bit of a chicken and egg situation and should be solved in the next week or two. Once 6.1 based packages are available we can finalise the testing.

BryceDFisher:
We have 6.1 images on the AKS side that we are testing with.

PR #54, opened 2023-10-10
Code Refactor 

stevendongatmsft:
In this PR, we made the skr and aasp compatible with each other. Each container can simply call SecureKeyRelease function to have keys automatically released without having to worry about authentication and retrieving UVM information. Also this PR prepares encfs sidecar containers for AKS.


cmd/aasp/main.go
```diff
@@ -59,10 +58,7 @@ var (
 )
 
 const (
-	aasp                       = "aasp"
-	AZURE_CLIENT_ID            = "AZURE_CLIENT_ID"
-	AZURE_TENANT_ID            = "AZURE_TENANT_ID"
-	AZURE_FEDERATED_TOKEN_FILE = "AZURE_FEDERATED_TOKEN_FILE"
+	aasp = "aasp"
```
hgarvison: Why is this lower case? General standards for constants is to have the names be upper case.

cmd/azmount/filemanager/azure.go
```diff
@@ -91,30 +93,44 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
-		// url Host denotes the scope/audience for which we need to get a token
-		logrus.Trace("Using token credentials to access private azure blob storage...")
-
-		var token common.TokenResponse
-		count := 0
-		logrus.Debugf("Getting token for https://%s", u.Host)
-		for {
-			token, err = common.GetToken("https://"+u.Host, identity)
-
+		ctx := context.TODO()
```
ksayid: Would it be worth adding a timeout to `ctx`?

pkg/skr/skr.go
```diff
@@ -88,30 +90,39 @@ func SecureKeyRelease(identity common.Identity, certState attest.CertState, SKRK
...
 	// retrieve an Azure authentication token for authenticating with AKV
 	if SKRKeyBlob.AKV.BearerToken == "" {
-		logrus.Info("Retrieving Azure authentication token...")
-		var ResourceIDTemplate string
-		// If endpoint contains managedhsm, request a token for managedhsm
-		// resource; otherwise for a vault
-		if strings.Contains(SKRKeyBlob.AKV.Endpoint, "managedhsm") {
-			logrus.Info("Requesting token for managedhsm...")
-			ResourceIDTemplate = ResourceIdManagedHSM
+		ctx := context.TODO()
```
ksayid: Same here. Would it be worth adding a timeout to `ctx`?

cmd/aasp/main.go
stevendongatmsft: done

pkg/skr/skr.go
stevendongatmsft: yup, this is on my to do list. It won't be just TODO()

stevendongatmsft: done

cmd/azmount/filemanager/azure.go
stevendongatmsft: done

cmd/remotefs/main.go
```diff
@@ -54,7 +54,7 @@ func usage() {
 
 func main() {
 	base64string := flag.String("base64", "", "base64-encoded json string with all information")
-	logLevel := flag.String("loglevel", "warning", "Logging Level: trace, debug, info, warning, error, fatal, panic.")
+	logLevel := flag.String("loglevel", "debug", "Logging Level: trace, debug, info, warning, error, fatal, panic.")
```
BryceDFisher: We would still want the default log level to be warning or error.  The defaults should be safe to use.  Debug can leak secrets which we wouldn't want by default.

cmd/aasp/main.go
```diff
@@ -425,16 +425,20 @@ func main() {
...
 
-	// pass in EncodedUvmInformation because ciruclar reference is created if we have the following func to retrieve THIM Cert
-	// Don't look at err here.  It will always be true because not all information is encoded in kata on AKS
 	EncodedUvmInformation, _ = common.GetUvmInformation()
 
-	thimCerts, err := azure_info.CertFetcher.GetThimCerts(azure_info.CertFetcher.Endpoint)
-	if err != nil {
-		logrus.Fatalf("Failed to retrieve thim certs: %s", err.Error())
+	if common.ThimCertsAbsent(&EncodedUvmInformation.InitialCerts) {
+		logrus.Info("ThimCerts is absent, retrieving THIMCerts.")
```
BryceDFisher: Retrieving THIMCerts from....?

cmd/aasp/main.go
```diff
@@ -425,16 +425,20 @@ func main() {
...
+		if err != nil {
+			logrus.Fatalf("Failed to retrieve thim certs: %s", err.Error())
+		}
+
+		EncodedUvmInformation.InitialCerts = *thimCerts
 	}
 
-	EncodedUvmInformation.InitialCerts = *thimCerts
+	//remove EncodedUvmReferenceInfo for now because MAA does not currently validate it
+	EncodedUvmInformation.EncodedUvmReferenceInfo = ""
```
BryceDFisher: Let's not hard remove this.  In my commits, its possible to set the path to the referenceinfo to something invalid and it will silently not include any reference info.  This will get caught as part of the attestation in MAA if the customer is expecting UVM info.  Also, tomorrow we're expecting a production signed UVM reference info, so it will all start working.

pkg/common/info.go
```diff
@@ -207,6 +207,10 @@ func GetReferenceInfoFile(securityContextDir string, ReferenceInfoFilename strin
 	return encodedUvmReferenceInfo, nil
 }
 
+func ThimCertsAbsent(thim *THIMCerts) bool {
+	return reflect.DeepEqual(*thim, THIMCerts{})
```
BryceDFisher: We don't need reflect.  Just check if the len of the vcek cert and certificate chain is 0.

cmd/aasp/main.go
```diff
@@ -262,22 +258,9 @@ func (s *server) UnWrapKey(c context.Context, grpcInput *keyprovider.KeyProvider
...
-			return nil, status.Errorf(codes.Internal, "Failed to obtain access token to MHSM: %v", err)
-		}
-	}
-
 	mhsm := skr.AKV{
-		Endpoint:    annotation.KmsEndpoint,
-		APIVersion:  "api-version=7.3-preview",
-		BearerToken: bearerToken,
+		Endpoint:   annotation.KmsEndpoint,
+		APIVersion: "api-version=7.3-preview",
```
BryceDFisher: Shouldn't we be using a newer production version?

cmd/aasp/main.go
```diff
@@ -360,7 +343,7 @@ func main() {
 	outfile := flag.String("outfile", "", "The file to save the wrapped data")
 	logLevel := flag.String("loglevel", "debug", "Logging Level: trace, debug, info, warning, error, fatal, panic.")
 
-	localhost := "localhost"
+	localhost := ""
```
BryceDFisher: We need to either put back localhost, or make it user configurable.  I did this to be able to query the endpoint from outside the pod, but that is super dangerous because it lets external entities use the attestation report with MAA.

cmd/skr/main.go
```diff
@@ -89,7 +90,7 @@ func main() {
 
 	EncodedUvmInformation, err := common.GetUvmInformation() // from the env.
 	if err != nil {
-		logrus.Fatalf("Failed to extract UVM_* environment variables: %s", err.Error())
+		logrus.Warnf("Failed to extract UVM_* environment variables: %s", err.Error())
```
BryceDFisher: Not sure if I made this warn or not, but let's update to Info.  In the AASP case, this is expected content.

cmd/skr/main.go
```diff
@@ -108,6 +109,21 @@ func main() {
...
+			logrus.Fatalf("Failed to retrieve thim certs: %s", err.Error())
+		}
+
+		EncodedUvmInformation.InitialCerts = *thimCerts
+	}
+
+	//temporary code. Remove uvm reference info if the running env is aks.
+	if msi.WorkloadIdentityEnabled() {
+		EncodedUvmInformation.EncodedUvmReferenceInfo = ""
+	}
```
BryceDFisher: I also don't want to hard remove uvm info here.  Either it exists or it doesn't.  For testing purposes until we have production signed uvm info, just set the path to the uvm info to an invalid path and don't fail.

examples/aasp/aasp-sample-template.yaml
```diff
@@ -20,6 +20,9 @@ spec:
     volumeMounts:
     - mountPath: /opt/confidential-containers/share/kata-containers/reference-info-base64
       name: endorsement-location
+    ports:
+      - containerPort: 8080
+      - containerPort: 50000
```
BryceDFisher: The default should not expose AASP outside of the pod.

pkg/msi/msi.go
```diff
@@ -3,18 +3,36 @@ package msi
...
+	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
 )
 
+const (
+	AZURE_CLIENT_ID                    = "AZURE_CLIENT_ID"
+	AZURE_TENANT_ID                    = "AZURE_TENANT_ID"
+	AZURE_FEDERATED_TOKEN_FILE         = "AZURE_FEDERATED_TOKEN_FILE"
+	WorkloadIdentityRquestTokenTimeout = time.Second * 7
```
BryceDFisher, line 17: How did you land on 7 seconds?

cmd/remotefs/main.go
stevendongatmsft: done

pkg/common/info.go
stevendongatmsft: done

pkg/msi/msi.go
stevendongatmsft: Really should've been 3-5 seconds because it should not take 7 seconds but considering occasional cluster network hiccup, I made it 7 seconds.

cmd/aasp/main.go
stevendongatmsft: yeah, should use api-version=7.4

examples/aasp/aasp-sample-template.yaml
stevendongatmsft: I'm not sure what needs to be done here.

BryceDFisher: Just remove the port declarations for the aasp container

stevendongatmsft: done

pkg/common/info.go
```diff
@@ -72,6 +71,10 @@ type UvmInformation struct {
 	EncodedUvmReferenceInfo string    // base64 encoded endorsements for the particular UVM image
 }
 
+// this will always be set in ACI by the contol plane and is optionally set in K8s.  Need to
+// use a default if the customer does not set
+const uvmSecurityCtxDirDefault = "/opt/confidential-containers/share/kata-containers"
```
SethHollandsworth, line 76: this var is only used in `GetUvmSecurityCtxDir` so we can probably make it a local variable there

cmd/aasp/main.go
```diff
@@ -262,22 +258,9 @@ func (s *server) UnWrapKey(c context.Context, grpcInput *keyprovider.KeyProvider
...
-	tenantID := os.Getenv(AZURE_TENANT_ID)
-	tokenFile := os.Getenv(AZURE_FEDERATED_TOKEN_FILE)
-	if clientID != "" && tenantID != "" && tokenFile != "" {
-		bearerToken, err = msi.GetAccessTokenFromFederatedToken(c, tokenFile, clientID, tenantID, "https://managedhsm.azure.net")
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "Failed to obtain access token to MHSM: %v", err)
-		}
-	}
-
 	mhsm := skr.AKV{
```
SethHollandsworth, line 261: does this not need a bearer token when using federated identities anymore?

cmd/aasp/main.go
```diff
@@ -469,6 +448,12 @@ func main() {
 		log.Printf("Warning: Env AZURE_CLIENT_ID is not set")
 	}
 
+	lis, err := net.Listen("tcp", *port)
+	if err != nil {
+		log.Fatalf("failed to listen on port %v: %v", *port, err)
+	}
+	log.Printf("Listening on port %v", *port)
+
 	url := *hostname + ":" + *httpport
```
SethHollandsworth, line 457: why not use `localhost` instead of `*hostname`?

cmd/azmount/filemanager/azure.go
```diff
@@ -91,30 +93,45 @@ func AzureSetup(urlString string, urlPrivate bool, identity common.Identity) err
...
-		// url Host denotes the scope/audience for which we need to get a token
-		logrus.Trace("Using token credentials to access private azure blob storage...")
-
-		var token common.TokenResponse
-		count := 0
-		logrus.Debugf("Getting token for https://%s", u.Host)
-		for {
-			token, err = common.GetToken("https://"+u.Host, identity)
-
+		ctx, cancel := context.WithTimeout(context.Background(), msi.WorkloadIdentityRquestTokenTimeout)
```
SethHollandsworth, line 96: typo in `WorkloadIdentityRquestTokenTimeout`, should be `WorkloadIdentityRequestTokenTimeout`. This is in 2 other spots too

pkg/common/http.go
```diff
@@ -58,16 +58,16 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
-
 	// Pull out response body
 	defer httpResponse.Body.Close()
 	httpResponseBodyBytes, err := io.ReadAll(httpResponse.Body)
 	if err != nil {
 		return nil, errors.Wrapf(err, "reading http response body failed")
 	}
 
-	return httpResponseBodyBytes, nil
+	if httpResponse.Status != "200 OK" && httpResponse.Status != "200 " {
```
SethHollandsworth, line 68: is moving this block down just to make sure the `defer` runs? it might be better to just move that line then

pkg/common/info.go
```diff
@@ -85,19 +88,27 @@ type UvmInformation struct {
...
 
 func GetUvmInformation() (UvmInformation, error) {
-	_, err := GetUvmSecurityCtxDir()
-	if err != nil {
-		return GetUvmInformationFromEnv()
-	} else {
-		return GetUvmInformationFromFiles()
+	contextDir, err := GetUvmSecurityCtxDir()
+	var info UvmInformation
+	if len(contextDir) > 0 {
```
SethHollandsworth, line 102: should this check `err != nil` rather than/in addition to the length of contextDir? it seems like that instance of err isn't referenced anywhere

pkg/common/http.go
stevendongatmsft: defer runs after this function exits. Moving this block does not affect defer.

cmd/azmount/filemanager/azure.go
stevendongatmsft: thanks. good catch

cmd/aasp/main.go
stevendongatmsft: this is moved to skr pkg skr.go#L108

pkg/common/info.go
stevendongatmsft: I would keep it a constant in case more files in the future gets added under this folder and we need other functions that references this const as well.

cmd/aasp/main.go
stevendongatmsft: done

pkg/common/info.go
BryceDFisher: I was actively ignoring the error.  Could use `_` I suppose.

stevendongatmsft: I think we can just leave it as is, otherwise we have to declare a `var err error` because we cannot use := on line 103) due to info

BryceDFisher (approved):

PR #56, opened 2023-10-18
add aks encfs readme instruction

stevendongatmsft:


examples/encfs/README.md
```diff
@@ -177,3 +178,223 @@ Alternatively, the whitelisted command in test-encfs-container outputs the follo
 This is a file inside the filesystem.
 This is a file inside the filesystem.
 ```
+### Step by Step Example On AKS
```
hgarvison, line 181: I feel like we should have two subfolders for the examples/encfs (and for skr in the future):

-- examples
  -- ENCFS
    -- ACI
    -- AKS
  -- SKR
     -- ACI
    -- AKS

This way you can separate the readmes (it's a long file now) and the required files for each deployment type

stevendongatmsft: If we do this, I feel the structure should be: 
--example
  --encfs
     --aci 
     --aks 
@BryceDFisher what's your opinion? I was investigating the tab thing you mentioned but apparently that is not supported in github.

hgarvison: Yes, I didn't include all the subdirs and it didn't get my spacing (example has encfs, skr, aasp, etc. and each of those has aci and aks)

BryceDFisher: If tabs work, let's do that.  If not, we should do subfolders.

BryceDFisher:
Closing this PR due to lack of support in AKS.

PR #59, opened 2023-10-25
removing hello world aks example

hgarvison:


SethHollandsworth (approved):
lgtm, thanks for adding both of the executables

PR #61, opened 2023-10-30
making import key tool work on Windows

hgarvison:
Passing Test Pipelines
ENCFS: https://github.com/microsoft/confidential-aci-examples/actions/runs/6734185413
SKR: https://github.com/microsoft/confidential-aci-examples/actions/runs/6734187200
Attestation: https://github.com/microsoft/confidential-aci-examples/actions/runs/6734190699

### DIFF
--- a/pkg/common/maa.go
+++ b/pkg/common/maa.go
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-package attest
+package common
 
 import (
 	"encoding/base64"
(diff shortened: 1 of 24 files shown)


pkg/common/maa.go
```diff
@@ -69,7 +68,7 @@ func newAttestSNPRequestBody(snpAttestationReport []byte, vcekCertChain []byte,
 	}
 	logrus.Debugf("base64urlEncodedUvmReferenceInfo: %s", base64urlEncodedUvmReferenceInfo)
 
-	if common.GenerateTestData {
+	if GenerateTestData {
```
SethHollandsworth, line 71: would it be better to make this an env var instead of an global shared variable?

SethHollandsworth (approved):
lgtm

pkg/common/maa.go
hgarvison: future work will resolve this

PR #64, opened 2023-11-03
fixing nonexistent paths in pipelines

hgarvison:
removing references to temporary /bin directory in the sidecar build pipelines


SethHollandsworth (approved):
lgtm

PR #65, opened 2023-11-03
revert hexstring.py removal change. 

stevendongatmsft:
Sacket complained the change broke customers' existing scripts. His blog also depends on this script. We need to provide backward compatibility. Bryce wanted us to put it back.

### DIFF
--- a/examples/encfs/generatefs/hexstring.py
+++ b/examples/encfs/generatefs/hexstring.py
@@ -1,0 +1,13 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import binascii
+import sys
+
+if len(sys.argv) != 2:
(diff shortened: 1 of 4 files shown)


SethHollandsworth (approved):

examples/encfs/generatefs/hexstring.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+import binascii
+import sys
+
+if len(sys.argv) != 2:
```
hgarvison, line 7: Just bringing the file back doesn't put the call to run it in the generatefs.sh script

stevendongatmsft:
ok, I tested both inputs of -kh and it works. I also tested creating new keys using importkey golang script and it works.

PR #66, opened 2023-11-06
apk upgrade for openssl

hgarvison:
Attempt 2 at fixing security issue with openssl

Passing pipelines:
ENCFS: https://github.com/microsoft/confidential-aci-examples/actions/runs/6773602758
SKR: https://github.com/microsoft/confidential-aci-examples/actions/runs/6773604681

### DIFF
--- a/docker/aasp/Dockerfile.build
+++ b/docker/aasp/Dockerfile.build
@@ -11,7 +11,7 @@ RUN cd cmd/aasp && CGO_ENABLED=0 GOOS=linux go build -o /aasp -ldflags="-s -w" m
 
 FROM alpine:3.18.4
 
-RUN apk update && apk --no-cache add curl
+RUN apk update && apk upgrade --no-cache && apk --no-cache add curl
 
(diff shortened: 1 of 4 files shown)


docker/aasp/Dockerfile.build
```diff
@@ -11,7 +11,7 @@ RUN cd cmd/aasp && CGO_ENABLED=0 GOOS=linux go build -o /aasp -ldflags="-s -w" m
 
 FROM alpine:3.18.4
 
-RUN apk update && apk --no-cache add curl
+RUN apk update && apk upgrade --no-cache && apk --no-cache add curl
```
BryceDFisher, line 14: Is this resolved by just using a newer alpine image instead?

hgarvison: 3.18.4 is the newest version (which is why I thought the previous PR resolved it) https://hub.docker.com/_/alpine/

SethHollandsworth (approved):
lgtm

had to double check the upgrade didn't need a `-y` like for apt: https://unix.stackexchange.com/questions/552809/automatic-yes-to-prompts-when-installing-package-on-alpine-linux

PR #70, opened 2023-11-15
include http response body in non 200 status error

stevendongatmsft:
Pipeline run: 
[SKR pipeline](https://github.com/microsoft/confidential-aci-examples/actions/runs/7116333295)
[ENCFS pipeline](https://github.com/microsoft/confidential-aci-examples/actions/runs/7106407253)

### DIFF
--- a/pkg/attest/attest_test.go
+++ b/pkg/attest/attest_test.go
@@ -82,7 +82,7 @@ func Test_MAA(t *testing.T) {
 		t.Fatalf("failed to deserialize attestation report")
 	}
 
-	ValidMAAEndpoint := "sharedeus.eus.attest.azure.net"
+	ValidMAAEndpoint := "sharedeus2.eus2.test.attest.azure.net"
 	ValidTEEType := "SevSnpVM"
 	ValidMAAAPIVersion := "api-version=2020-10-01"
 
@@ -143,8 +143,7 @@ func Test_MAA(t *testing.T) {
 		keyBlob              []byte
 		uvmReferenceInfo     []byte
 
-		expectedError error
-		expectErr     bool
+		expectErr bool
 	}
 
 	testcases := []*testcase{
@@ -165,8 +164,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              runtimedata_json_bin,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: nil,
-			expectErr:     false,
+			expectErr: false,
 		},
 		// MAA_InvalidCertChain testing passes if the MAA responds with Bad Request
 		{
@@ -184,8 +182,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              TestKeyBlobBytes,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 		// MAA_InvalidKeyBlob testing passes if the MAA responds with Bad Request as the key blob does not match the one in the attestation report
 		{
@@ -203,8 +200,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              make([]byte, 4),
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 		// MAA_TamperedKeyBlob testing passes if the MAA responds with Bad Request as the key blob does not match the one in the attestation report
 		{
@@ -222,8 +218,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              TestKeyBlobBytesTampered,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 		// MAA_InvalidPolicyBlob testing passes if the MAA responds with Bad Request as the policy does not match the one in the attestation report
 		{
@@ -241,8 +236,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              TestKeyBlobBytes,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 
 		// MAA_TamperedPolicyBlob testing passes if the MAA responds with Bad Request as the policy does not match the one in the attestation report
@@ -261,8 +255,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              TestKeyBlobBytes,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 		{
 			name: "MAA_CorruptedSNPReport",
@@ -279,8 +272,7 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              ProductionTestKeyBlobBytes,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectErr: true,
 		},
 		{
 			name: "MAA_CorruptedSigSNPReport",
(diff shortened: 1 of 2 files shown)


pkg/attest/attest_test.go
```diff
@@ -297,8 +305,9 @@ func Test_MAA(t *testing.T) {
 			keyBlob:              ProductionTestKeyBlobBytes,
 			uvmReferenceInfo:     uvm_reference_info_bin,
 
-			expectedError: errors.New("pulling maa post response failed: http response status equal to 400 Bad Request"),
-			expectErr:     true,
+			expectedError:   errors.New("http response status equal to 400 Bad Request"),
```
BryceDFisher: Rather than keeping strings in sync everywhere, this is an example of when its useful to have typed errors that you can check.  That way the string itself can be anything.  Something like 
```go
type HTTPError string
func (h HTTPError) Error() string{
 return "http response status equal to " + string(HTTPError)
}
```

Then you could check it in the tests with `if errors.Is(HTTPError){...}` which will also deal with unwrapping the error if it gets wrapped anywhere.

pkg/common/http.go
```diff
@@ -66,7 +66,7 @@ func HTTPResponseBody(httpResponse *http.Response) ([]byte, error) {
 	}
 
 	if httpResponse.Status != "200 OK" && httpResponse.Status != "200 " {
-		err = errors.Errorf("http response status equal to %s", httpResponse.Status)
+		err = errors.Wrapf(errors.Errorf("http response status equal to %s", httpResponse.Status), string(httpResponseBodyBytes))
```
BryceDFisher: You would also likely want to include the bytes in the other return above.  It's possible to get a partial read before the connection is unexpectedly closed which could provide insight to the issue.

stevendongatmsft: I ported the changes you made in demos repo to here so that the partial read before connection close won't be an issue.

pkg/common/http.go
```diff
@@ -59,15 +59,18 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
 	// Pull out response body
-	defer httpResponse.Body.Close()
-	httpResponseBodyBytes, err := io.ReadAll(httpResponse.Body)
-	if err != nil {
-		return nil, errors.Wrapf(err, "reading http response body failed")
+	var httpResponseBodyBytes []byte
+	if httpResponse != nil && httpResponse.Body != nil {
+		defer httpResponse.Body.Close()
+		limitSize := httpResponse.ContentLength
+		const mb134 = 1 << 27 //134MB
```
hgarvison: Can you come up with a more descriptive name or add a comment that explains why it's 134MB?

pkg/common/http.go
```diff
@@ -59,15 +59,18 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
+		if limitSize == 0 || limitSize > mb134 {
+			limitSize = mb134
```
hgarvison: Since 134MB is the limit, the constant (mb134) should be called "respLenLimit134MB" and limitSize should be responseLength (respLen or something to that effect).

pkg/common/http.go
```diff
@@ -59,15 +59,18 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
+		}
+		httpResponseBodyBytes, _ = io.ReadAll(io.LimitReader(httpResponse.Body, int64(limitSize)))
```
hgarvison: Why do we need to do all this instead of just reading the response body?

pkg/attest/attest_test.go
```diff
@@ -1,340 +1,342 @@
-// Copyright (c) Microsoft Corporation.
-// Licensed under the MIT License.
-
-package attest
```
hgarvison: Why does it look like the whole file was changed?

pkg/common/http.go
BryceDFisher: It just seemed like an appropriate size that was an even bit shift.

pkg/common/http.go
BryceDFisher: What if we connect to a malicious endpoint that instead of being reasonable feeds us an unlimited response length try to cause buffer overflows?

pkg/attest/attest_test.go
BryceDFisher: This is probably a line ending encoding issue.  @stevendongatmsft please double check this.

pkg/common/http.go
hgarvison: Then we should make a comment that it's arbitrary so that in the future no one has to guess why.

pkg/common/http.go
hgarvison: Then we should make a comment that that's why we're doing this and probably put it in the description of the PR...

pkg/attest/attest_test.go
stevendongatmsft: done

pkg/attest/attest_test.go
stevendongatmsft: this is done

pkg/common/http.go
```diff
@@ -58,19 +66,20 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
+	// Pull out response body. We are using a LimitReader to prevent unlimited server response causing buffer overflow
 	var httpResponseBodyBytes []byte
 	if httpResponse != nil && httpResponse.Body != nil {
 		defer httpResponse.Body.Close()
-		limitSize := httpResponse.ContentLength
-		const mb134 = 1 << 27 //134MB
-		if limitSize == 0 || limitSize > mb134 {
-			limitSize = mb134
+		respLen := httpResponse.ContentLength
+		//134MB is an arbitraty limit size that is appropriate for http reponse using bit manipulation
```
hgarvison: arbitrary

hgarvison (approved):

pkg/common/http.go
```diff
@@ -58,16 +66,20 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
-	defer httpResponse.Body.Close()
-	httpResponseBodyBytes, err := io.ReadAll(httpResponse.Body)
-	if err != nil {
-		return nil, errors.Wrapf(err, "reading http response body failed")
+	// Pull out response body. We are using a LimitReader to prevent unlimited server response causing buffer overflow
+	var httpResponseBodyBytes []byte
+	if httpResponse != nil && httpResponse.Body != nil {
+		defer httpResponse.Body.Close()
+		respLen := httpResponse.ContentLength
+		//134MB is an arbitrary limit size that is appropriate for http reponse using bit manipulation
```
KenGordon, line 74: 134? 128 surely? And when can this happen? Do we have 128MB to use as a buffer? These objects are all in the 10s of K.

pkg/common/http.go
```diff
@@ -58,16 +66,20 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
...
+		if respLen < 1 || respLen > respLenLimit134mb {
+			respLen = respLenLimit134mb
+		}
+		httpResponseBodyBytes, _ = io.ReadAll(io.LimitReader(httpResponse.Body, int64(respLen)))
 	}
-
-	if httpResponse.Status != "200 OK" && httpResponse.Status != "200 " {
-		err = errors.Errorf("http response status equal to %s", httpResponse.Status)
+	if httpResponse.StatusCode < 200 || httpResponse.StatusCode > 207 {
+		return nil, errors.Wrapf(&HTTPError{httpResponse.Status}, string(httpResponseBodyBytes))
```
KenGordon, line 82: Does this response come from the wire? Are we signing up to take up to 128MB of stuff from some external http source and make a string out of it and pass that back up the stack?

pkg/common/http.go
stevendongatmsft: By "these object", are you referring to the http response? Just curious, even if the objects are all in the 10s of K, would it matter if the reading limit we set is not in 10s of K? 
Bit manipulation of 2^27 just seems to be an appropriate number for most of the http response size and does not require that much computation.

pkg/common/http.go
stevendongatmsft: These are the http responses, the responses could be from maa endpoint, mhsm endpoint or authentication endpoint. Since we are not responding to those endpoints, I don't see any issues here. 

Also the limit reader is just to guarantee that we will never get a buffer overflow, most of the responses are several bytes at most in practice.

pkg/common/http.go
KenGordon: The machine might only have 1GB of RAM, to contain an OS, some other process etc. If you are fooled by a malicious fake MAA into allocating and populating 128MB of memory when the most you might every need is a few 10s of k then I think you are more vulnerable compared to limiting it to something more sensible, even 1MB.

2^27 is a constant. Writing it as 2^27, 2*2*2...*2, 128*1024*1024 or in long hand it takes the same time to run. I don;t care about that, I care that you are allocating a really big lump of memory if the outside world ask you to.

stevendongatmsft: So what do you think is a reasonable memory limit?

KenGordon: Is this in fact an error message destined to be shown to the user? If so, 1) check with Greg, 2) guess something like 1k, maybe 10k if it might include a bunch of nested error messages generated in the confusing style of all these services.

BryceDFisher: I've seen these be pretty big but I haven't measured.  Maybe 1mb is a good compromise and we can move forward with the PR.

Yes, you must read this response body to get any sort of trouble shooting information from MAA and from AKV.

stevendongatmsft: I changed it to 2^22 roughly 4 megabytes. Manually tested and it passed encfs pipeline.

stevendongatmsft: changed to 2^20 roughly 1mb and both manual testing and pipeline passed. I'm going to rebase and merge

PR #72, opened 2023-11-21
Change skr base image to Mariner

SethHollandsworth:
Changing base image for SKR image to Mariner to be in line with Azure updates. 

Putting in draft mode for now until the kernel is patched in ACI to allow for exec-ing into this image

### DIFF
--- a/pkg/common/maa.go
+++ b/pkg/common/maa.go
@@ -174,7 +174,7 @@ func (maa MAA) Attest(SNPReportHexBytes []byte, vcekCertChain []byte, policyBlob
 	if err != nil {
 		// added debug information if needed
 		if httpResponse != nil && httpResponse.Header != nil && httpResponse.Header.Get("x-ms-request-id") != "" {
-			logrus.Debugf("MAA Response header: %s", httpResponse.Header.Get("x-ms-request-id"))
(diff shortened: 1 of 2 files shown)


pkg/common/maa.go
```diff
@@ -172,6 +172,10 @@ func (maa MAA) Attest(SNPReportHexBytes []byte, vcekCertChain []byte, policyBlob
 	logrus.Debugf("Posting MAA Attestation Request to %s", uri)
 	httpResponse, err := HTTPPRequest("POST", uri, maaRequestJSONData, "")
 	if err != nil {
+		// added debug information if needed
+		if httpResponse != nil && httpResponse.Header != nil && httpResponse.Header.Get("x-ms-request-id") != "" {
+			logrus.Debugf("MAA Response header: %s", httpResponse.Header.Get("x-ms-request-id"))
```
stevendongatmsft: I would make this message more specific to `x-ms-request-id`, maa request http response could have multiple headers, this is just one of them so that it's easier to debug for customers. I would also consider making this an `info` instead of a debug so that customers do not have to re-run the container to get the request id(if most customers do not run it as debug by default)

SethHollandsworth:
closing and will open a new PR once the kernel is patched in ACI

PR #76, opened 2023-12-01
Making grpc server optional

SethHollandsworth:
Having the gRPC server on by default on port 50000 might interfere with customer servers, so we should turn it off by default and enable it by setting the KEYPROVIDER_SOCK environment variable.

This was tested in the ACI and AKS pipelines manually and will make respective PRs once there is an official build. Thanks!

### DIFF
--- a/cmd/skr/main.go
+++ b/cmd/skr/main.go
@@ -30,18 +30,14 @@ func usage() {
 	flag.PrintDefaults()
 }
 
-// TODO:
-// - vcek source is different
-// - aci gets security policy and hashes it
-
 func main() {
 	azureInfoBase64string := flag.String("base64", "", "optional base64-encoded json string with azure information")
 	logLevel := flag.String("loglevel", "warning", "Logging Level: trace, debug, info, warning, error, fatal, panic.")
 	logFile := flag.String("logfile", "", "Logging Target: An optional file name/path. Omit for console output.")
-	httpPort := flag.String("port", "8080", "Port on which to listen")
+	port := flag.String("port", "8080", "Port on which to listen")
 	allowTestingMismatchedTCB := flag.Bool("allowTestingMismatchedTCB", false, "For TESTING purposes only. Corrupts the TCB value")
 	// NOTE: these 4 input arguments are typically only used in AKS, not ACI
-	grpcPort := flag.String("keyprovider_sock", "127.0.0.1:50000", "Port on which the grpc key provider to listen")
+	serverType := flag.String("server_type", "http", "Choose whether to use http or grpc for the server type")
 	infile := flag.String("infile", "", "The file with its content to be wrapped")
 	key_path := flag.String("keypath", "", "The path to the wrapping key")
 	outfile := flag.String("outfile", "", "The file to save the wrapped data")
@@ -85,7 +81,7 @@ func main() {
 	logrus.Infof("Args:")
 	logrus.Infof("   Log Level:     %s", *logLevel)
 	logrus.Infof("   Log File:      %s", *logFile)
-	logrus.Debugf("   Port:          %s", *httpPort)
+	logrus.Debugf("   Port:          %s", *port)
 	logrus.Debugf("   Hostname:      %s", *hostname)
 	logrus.Debugf("   azure info:    %s", *azureInfoBase64string)
 	logrus.Debugf("   corrupt tcbm:  %t", *allowTestingMismatchedTCB)
@@ -151,7 +147,7 @@ func main() {
 	}
 
 	// See above comment about hostname and risk of breaking confidentiality
-	url := *hostname + ":" + *httpPort
+	url := *hostname + ":" + *port
 
 	logrus.Trace("Getting initial TCBM value...")
 	var tcbm string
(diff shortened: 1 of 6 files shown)


cmd/skr/main.go
```diff
@@ -41,7 +41,7 @@ func main() {
 	httpPort := flag.String("port", "8080", "Port on which to listen")
 	allowTestingMismatchedTCB := flag.Bool("allowTestingMismatchedTCB", false, "For TESTING purposes only. Corrupts the TCB value")
 	// NOTE: these 4 input arguments are typically only used in AKS, not ACI
-	grpcPort := flag.String("keyprovider_sock", "127.0.0.1:50000", "Port on which the grpc key provider to listen")
+	grpcPort := flag.String("keyprovider_sock", os.Getenv("KEYPROVIDER_SOCK"), "Port on which the grpc key provider to listen")
```
hgarvison: to match the other flags, I would make this camel-case

cmd/skr/main.go
```diff
@@ -177,19 +177,21 @@ func main() {
...
-	log.Printf("Listening on port %v", *grpcPort)
-	//start grpc server
-	s := grpc.NewServer()
-	server := server.Server{ServerCertState: &certState, EncodedUvmInformation: &EncodedUvmInformation, Azure_info: &info}
-	keyprovider.RegisterKeyProviderServiceServer(s, &server)
-	reflection.Register(s)
-	log.Printf("server listening at %v", lis.Addr())
-	if err := s.Serve(lis); err != nil {
-		log.Fatalf("failed to start GRPC server: %v", err)
+	if *grpcPort == "" {
```
hgarvison: Should be != ""?

cmd/skr/main.go
SethHollandsworth: I'll ask and see, env vars are usually capital snake case. I feel like the other ones should be changed too

hgarvison (approved):

cmd/skr/main.go
KenGordon: This has a host address as well as the port. If we don’t enforce exactly over that then the host could be set to 0.0.0.0 and then be available outside of the UVM. If you look at the comments (approx line 88) about taking the host as part of the command line you will see that we previously decided that that was too much of a risk. If people want to do that they will have to make the change there and build a bespoke container.

Also, the passing of this as an individual env variable directly to the binary rather than to the skr.sh seems to be a mismatch of style. That leads to another question about whether it is ever that case that both the http and grpc endpoints will be in use at once. If not, then maybe better to use the existing port variable and signal the grpc vs http by some other means.

KenGordon (commented):
I’d prefer the http vs grpc thing to be a switch. I think as this stands that there must always be an http server even if it is never used.

Too late I guess, but by combining this way the size and TCB is increased.

cmd/skr/main.go
SethHollandsworth: I agree with the first paragraph, I'll make changes so KEYPROVIDER_SOCK only takes in a port instead of an address and port.

For the styling, I'm working on a PR that will let users use capital snake case e.g. LOG_FILE (which is more common for env vars) instead of the less conventional casing of LogFile. It'll still have backwards compatibility but will let users directly use `/bin/skr` more easily, instead of having to go through `skr.sh`

KenGordon: Think about how this will be for a user. There are a number of flags that apparently do the same thing, ie set the port for the endpoint that provides the key. How do they choose? RTFM vs blindingly obvious tells me that the port should be port and that switching between GRPC and HTTP should not be a side effect of setting the port.

I am also not entirely convinced that one container ought to handle both. This expands the TCB for both use cases but I will let that go for the moment. People caring strongly about TCB will take the code and adapt. That brings me back to the env variable style and which layer ought to be responsible. It is too late to change style. Even if the existing style is "wrong" it is the actual style deployed. You cannot change that without breaking people and so introducing a different style means two styles will be in place and that will be confusing.

The purpose of the skr.sh is to adapt the skr binary to a sidecar. Some users will not want to have a sidecar but instead bundle the binary into their own container. Thus that skr.sh always has a place and the layering is a deliberate splitting of responsibilities.

hgarvison (approved):

docker/skr/skr-debug.sh
```diff
@@ -39,6 +39,22 @@ if [ -n "${LogLevel}" ]; then
   CmdlineArgs="${CmdlineArgs} -loglevel ${LogLevel}"
 fi
 
+if [ -n "${KeyproviderSock}" ]; then
+  CmdlineArgs="${CmdlineArgs} -keyprovider_sock ${KeyproviderSock}"
+fi
+
+if [ -n "${InFile}" ]; then
```
hgarvison: These last 3 vars aren't used for the grpc server. They're used for using SKR to create "wrapped data" (ie encrypting stuff) so they wouldn't normally be used when SKR is running in ACI/AKS.

hgarvison (changes requested):
Remove extra vars from shell scripts per other comment

hgarvison (approved):

SethHollandsworth:
> I’d prefer the http vs grpc thing to be a switch. I think as this stands that there must always be an http server even if it is never used.
> 
> Too late I guess, but by combining this way the size and TCB is increased.

Updated now so the http/grpc option is a switch. If KeyproviderSock is provided as an env var, the container will use grpc. Otherwise and by default it will use http. Test runs are on the way

KenGordon:
> > I’d prefer the http vs grpc thing to be a switch. I think as this stands that there must always be an http server even if it is never used.
> > Too late I guess, but by combining this way the size and TCB is increased.
> 
> Updated now so the http/grpc option is a switch. If KeyproviderSock is provided as an env var, the container will use grpc. Otherwise and by default it will use http. Test runs are on the way
Can I repeat my previous concern "RTFM vs blindingly obvious tells me that the port should be port and that switching between GRPC and HTTP should not be a side effect of setting the port."

hgarvison:
SKR Pipelines: https://github.com/microsoft/confidential-aci-examples/actions/runs/7573419398

SethHollandsworth:
Mariner Pipeline: https://dev.azure.com/mariner-org/mariner/_build/results?buildId=485505&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=07ba3131-9dbf-5d34-279b-d3a6aa521384

cmd/skr/main.go
SethHollandsworth: Updated so env var casing is staying consistent with what is already present in the repo. There is no longer plans for another PR to change the casing for existing environment variables

KenGordon (approved):
This looks good.

PR #78, opened 2023-12-05
Build SKR tests with cover enabled

BryceDFisher:
Enable code coverage profiles for the skr test containers.


hgarvison:
This covers the tests being run by the attestation container dockerfile, but it doesn't cover the grpc_test in the skr dir (gets built in the skr dockerfile) or the cache_test in cmd/azmount/filemanager (built in the encfs build script which could/should be modified to match the format of the other dockerfiles). If this is the intended effect with this PR then I'm okay with it.

BryceDFisher:
This was just adding coverage to the tests already being run.  Future tests or tests in other areas also need to run coverage.

hgarvison (approved):

PR #80, opened 2023-12-06
make repository dispatch call to run aci-examples tests

hgarvison:
Make repository dispatch call to aci-examples repo to run tests for sidecars automatically.

Review in conjunction with: https://github.com/microsoft/confidential-aci-examples/pull/116

### DIFF
--- a/.github/workflows/push_attestation_image.yml
+++ b/.github/workflows/push_attestation_image.yml
@@ -62,3 +62,15 @@ jobs:
           push: true
           tags: |
             ${{ secrets.REGISTRY_NAME }}.${{ secrets.REGISTRY_DOMAIN }}/attestation:${{ steps.get_image_tag.outputs.image_tag }}
+
+      - name: Run Attestation Tests
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          curl --fail-with-body -L \
+          -X POST \
+          -H "Accept: application/vnd.github+json" \
+          -H "Authorization: Bearer ${{ secrets.ACC_REPO_DISPATCH_TOKEN }}" \
+          -H "X-GitHub-Api-Version: 2022-11-28" \
+          https://api.github.com/repos/microsoft/confidential-aci-examples/dispatches \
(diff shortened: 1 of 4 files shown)


.github/workflows/push_attestation_image.yml
```diff
@@ -62,3 +62,15 @@ jobs:
           push: true
           tags: |
             ${{ secrets.REGISTRY_NAME }}.${{ secrets.REGISTRY_DOMAIN }}/attestation:${{ steps.get_image_tag.outputs.image_tag }}
+
+      - name: Run Attestation Tests
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          curl -L \
```
ksayid: should a `--fail` or `--fail-with-body` flag be added to the `curl` command? I think `curl` will treat an HTTP response with a 4xx or 5xx status code as successful. Adding the flags would make `curl` return a non-zero exit code. 

I don't think we'd want the workflow to proceed if the tests haven't been started.

.github/workflows/push_encfs_image.yml
```diff
@@ -57,8 +58,24 @@ jobs:
...
             ${{ secrets.REGISTRY_DOMAIN }} \
-            encfs:$image_tag \
+            encfs:${{ steps.get_image_tag.outputs.image_tag }} \
             --skip-login
+
+      - name: Run ENCFS Tests
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          curl -L \
```
ksayid: same here?

.github/workflows/push_skr_image.yml
```diff
@@ -56,8 +57,24 @@ jobs:
...
             ${{ secrets.REGISTRY_DOMAIN }} \
-            skr:$image_tag \
+            skr:${{ steps.get_image_tag.outputs.image_tag }} \
             --skip-login
+
+      - name: Run SKR Tests
+        env:
+          PR_NUMBER: ${{ github.event.number }}
+        run: |
+          curl -L \
```
ksayid: and here?

.github/workflows/push_attestation_image.yml
hgarvison: That works for me. I wasn't aware of that curl flag.

ksayid (approved):
LGTM

PR #85, opened 2024-01-04
update skr aks readme to describe key info file

hgarvison:
Updating SKR AKS readme and scripts to describe the key-info.json file created

### DIFF
--- a/examples/skr/aks/README.md
+++ b/examples/skr/aks/README.md
@@ -17,6 +17,8 @@
 This guide provides instructions on how to perform secret provisioning in AKS using the SKR container.
 By following these instructions, users can ensure that their secrets are securely stored and can be used by their applications.
 
+Note: These instructions assume the usage of an Azure Key Vault and Microsoft Azure Attestation Endpoint, however the steps can be adapted for other key vaults and attestation services.
+
 ## Export Environment Variables
 
 ```bash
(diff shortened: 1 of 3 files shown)


examples/skr/aks/README.md
```diff
@@ -44,9 +44,9 @@ You can follow the installation instructions for Azure CLI [here](https://learn.
 
 ### 2. Obtain An Attestation Endpoint
 
-Below are the MAA endpoints for the four regions Confidential Containers in AKS is currently available in. 
+Below are the MAA endpoints for the four regions Confidential Containers in AKS is currently available in.
```
SethHollandsworth: nit: end of sentence preposition, also "in" to "on"

Below are the MAA endpoints for the four regions in which  Confidential Containers on AKS is currently available

examples/skr/aks/README.md
```diff
@@ -56,9 +56,22 @@ West Europe: sharedweu.weu.attest.azure.net
 This is a one time effort.
 Once the key is created, it can be used to protect as many secrets as desired.
 
-The `setup-key-mhsm.sh` script creates a private/public key pair in the keyvault and a key release policy with the file name `keyname-release-policy.json`.
-The public key is then downloaded as `keyname-pub.pem` and a key info file is generated.
+The `setup-key-mhsm.sh` script creates a private/public key pair in the keyvault and a key release policy with the file name `<keyname>-release-policy.json`.
```
SethHollandsworth, line 61: since the json is labeled "kms_endpoint", is it worth saying this setup is for Azure mHSM but AKV and other KMS can also be used?

hgarvison: Yeah, that's my issue with us not being consistent on the naming. All of our setup is for AKV/MAA and so the names in the files are for that (with the exception of the AASP code), but I was under the impression that other kms/attestation endpoints were usable, so it makes more sense to update variables to be generic and specify that the setup scripts assume the usage of Azure kms/attestation.

SethHollandsworth (approved):
lgtm

examples/skr/aks/README.md
```diff
@@ -44,9 +46,9 @@ You can follow the installation instructions for Azure CLI [here](https://learn.
 
 ### 2. Obtain An Attestation Endpoint
 
-Below are the MAA endpoints for the four regions Confidential Containers in AKS is currently available in. 
+Below are the MAA endpoints for the four regions in which Confidential Containers on AKS is currently available.
 
-East US: sharedeus.eus.attest.azure.net
```
SethHollandsworth, line 49: nit: could we put line breaks between these 4 so they're easier to read

PR #87, opened 2024-01-09
remove aasp references

hgarvison:
Remove remaining AASP references to be consistent in naming throughout

Pipelines:
ENCFS: https://github.com/microsoft/confidential-aci-examples/actions/runs/7463714326
SKR: https://github.com/microsoft/confidential-aci-examples/actions/runs/7463717634

### DIFF
--- a/examples/skr/aks/unwrap.sh
+++ b/examples/skr/aks/unwrap.sh
@@ -20,7 +20,7 @@ if [ -z "${KEY_PROVIDER_PORT}" ]; then
   KEY_PROVIDER_PORT=50000
 fi
 
-AAA=`printf aasp | base64 -w0`
+AAA=`printf skr | base64 -w0`
 ANNO=`cat ${infile}`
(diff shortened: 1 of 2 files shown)


examples/skr/aks/unwrap.sh
```diff
@@ -20,7 +20,7 @@ if [ -z "${KEY_PROVIDER_PORT}" ]; then
   KEY_PROVIDER_PORT=50000
 fi
 
-AAA=`printf aasp | base64 -w0`
```
SethHollandsworth, line 23: Would doing this break things if they uploaded a key with the old version of the file then updated the container and tried to release the same key?

hgarvison: It wouldn't break key release since the Attestation Agent value doesn't have to do with the upload/claims (see the mhsm script in the same dir), but it would break the call to the grpc endpoint to initiate key release. I made the appropriate changes for that, so as long as customers use the updated skr, instead of the old one they would be fine. 

Also, the use of AASP is AKS SKR specific and as far as I know, not many customers are onboarded to it yet. So, it wouldn't break anyone using SKR for ACI. I figured we'd make this change now to avoid breaking more later and avoid the question of "what is AASP."

PR #90, opened 2024-01-16
update key vault role name

hgarvison:
Update the key vault role name required for the managed id to Key Vault Crypto Service Release User


examples/encfs/README.md
```diff
@@ -42,7 +42,7 @@ Copy the AttestURI endpoint value (sans https://) to the [Attestation Authority
 The user needs to instantiate an Azure Key Vault resource that supports storing keys in an mHSM: a [Premium vault](https://learn.microsoft.com/en-us/azure/key-vault/general/overview) or an [mHSM resource](https://docs.microsoft.com/en-us/azure/key-vault/managed-hsm/overview).
 
 After setting up an Azure Key Vault resource, generate a user-assigned managed identity that will be attached to the container group so that the containers have the correct access permissions to Azure services and resources.
-The managed identity needs *Key Vault Crypto Officer* and *Key Vault Crypto User* roles if using AKV key vault or *Managed HSM Crypto Officer* and *Managed HSM Crypto User* roles for /keys if using AKV managed HSM.
+The managed identity needs *Key Vault Crypto Service Release User* role if using AKV key vault or *Managed HSM Crypto Officer* and *Managed HSM Crypto User* roles for /keys if using AKV managed HSM.
```
KenGordon: Is the new role only for AKV and not mHSM? Also I think you should mention the previous required roles so people understand that that will still work and not be randomly assigning roles in the case of an issue unrelated to roles.

KenGordon (changes requested):

examples/encfs/README.md
hgarvison: @BryceDFisher said it was only for AKV and the mHSM roles didn't change. 

I'll update the text to reference the previous roles.

BryceDFisher: The roles should be:
AKV: Key Vault Crypto Service Release User
mHSM: Managed HSM Crypto Service Release User

These both supersede requiring Officer or User roles.

BryceDFisher: Please make sure the mhsm references match the above as well.

BryceDFisher (approved):

KenGordon (approved):

PR #96, opened 2024-02-12
Adding Vulnerability Scan

hgarvison:
Adding vulnerability scanning to main pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ on:
     # The branches below must be a subset of the branches above
     branches: [ "main" ]
   schedule:
-    - cron: '20 20 * * 6'
+    - cron: '0 0 * * *'
 
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -71,6 +71,21 @@ jobs:
...
 
+    - name: Aqua Security Trivy
+      uses: aquasecurity/trivy-action@0.17.0
+      with:
+        scan-type: 'fs'
+        scan-ref: '.'
+        format: 'sarif'
+        output: 'trivy-results.sarif'
+        exit-code: 1
+        severity: UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL
```
SethHollandsworth, line 82: Do we want to error out if there's a low severity action found? It seems like that would happen often and may not be resolvable at any given moment

hgarvison: As long as it still notifies us of the low severity action, I'm okay with removing it from this list.

SethHollandsworth: it looks like there isn't a good way to only exit on High and Critical while still alerting on the lower severities. We could have two separate steps, one to make a report with all vulnerabilities and one that exits on severe vulnerabilities. 

But for now, I think we can trial having all vulnerabilities error out and see how it goes

ksayid (commented):
Since vulnerability scanning is being added, should we change the cron job frequency? Right now, I think it's once a week.

hgarvison:
> Since vulnerability scanning is being added, should we change the cron job frequency? Right now, I think it's once a week.

@ksayid are you thinking a daily scan?

ksayid:
@hgarvison Yeah, I was thinking daily, but either would be ok I think. Like the alert level comment, maybe we can trial with daily and see how it goes?

ksayid (approved):

PR #100, opened 2024-02-22
unmount remote file system folders before terminate

stevendongatmsft:


docker/encfs/encfs.sh
```diff
@@ -22,11 +22,56 @@ fi
 
 echo CmdlineArgs = $CmdlineArgs
 
+cleanup_unmount() {
+  mount_array=$(process_encoded_string $EncfsSideCarArgs)
+  echo "Cleaning up by unmounting the following remote filesystem folder."
+  for mount_point in "${mount_array[@]}"; do
+    echo "attempting to unmount $mount_point"
+    output=$(umount $mount_point 2>&1)
```
hgarvison: why not `umount $mount_point 2>&1` and then just check the exit status `if [ $? -eq 0 ]`? If the first command has any output, it should automatically print it out I would think and then there's no need for extra variable assignments...

docker/encfs/encfs.sh
```diff
@@ -22,11 +22,56 @@ fi
...
+    exit_status=$?
+    echo "$output"
+    if [ $exit_status -eq 0 ]; then
+      echo "Unmounting was successful."
+    else
+      echo "Unmounting failed with exit status $exit_status."
```
hgarvison, line 34: Should we not terminate here?

docker/encfs/encfs.sh
```diff
@@ -22,11 +22,56 @@ fi
...
+            for folder in "$mount_point"/.*; do
+                if [[ -d "$folder" && $folder =~ "$mount_point"/.filesystem-[0-99]* ]]; then
+                  set["$folder"]=1
+                fi
+            done
+        done
+        keys=("${!set[@]}")
+        echo "${keys[@]}"
+    else
+        echo "Error: The decoded string is not a valid JSON object."
```
hgarvison, line 61: Should we not exit here?

docker/encfs/encfs.sh
stevendongatmsft: So if unmounting the first mounted folder failed, we should continue to try to unmount other folders.

docker/encfs/encfs.sh
stevendongatmsft: yes

hgarvison (approved):

PR #108, opened 2024-03-22
Update encfs dockerfile

hgarvison:
update encfs dockerfile to include build steps like skr

### DIFF
--- a/docker/encfs/Dockerfile.encfs
+++ b/docker/encfs/Dockerfile.encfs
@@ -1,8 +1,19 @@
+FROM mcr.microsoft.com/devcontainers/go:dev-1.22-bookworm AS build
+ENV PROJECT_DIR=/go/src/github.com/microsoft
+WORKDIR ${PROJECT_DIR}/confidential-sidecar-containers
+COPY . ./
+RUN cd tools/get-snp-report && make && mv bin/get-snp-report /
+
+RUN CGO_ENABLED=0 GOOS=linux go build -o / ./cmd/azmount
+RUN CGO_ENABLED=0 GOOS=linux go build -o / ./cmd/remotefs
+
 FROM alpine:3.18.6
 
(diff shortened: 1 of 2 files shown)


docker/encfs/build.sh
```diff
@@ -8,23 +8,7 @@ pushd $(dirname $0)
...
-popd
-cp ../../tools/get-snp-report/bin/get-snp-report ./bin
-cp ../../tools/get-snp-report/bin/get-fake-snp-report ./bin
-
-docker build --tag encfs -f Dockerfile.encfs .
+docker build --tag encfs -f Dockerfile.encfs ../..
 
 # clean up
-rm -rf bin
 popd
```
ksayid, line 14: Maybe `popd` can also be removed since `pushd` isn't used anymore?

docker/encfs/Dockerfile.encfs
```diff
@@ -1,8 +1,19 @@
+FROM mcr.microsoft.com/devcontainers/go:dev-1.22-bookworm AS build
```
ksayid, line 1: Just wondering, when do we want to use a dev container image? 
Currently, I think the repo is using two types of go image (both with 1.22.1 available)
> mcr.microsoft.com/oss/go/microsoft/
> mcr.microsoft.com/devcontainers/go

docker/encfs/build.sh
hgarvison: pushd is still on line 7...

docker/encfs/Dockerfile.encfs
hgarvison: The dev image has make (amongst other things) in it, so we're using it for the build of SKR and ENCFS. As for the oss/go/microsoft image, we're using that in the attestation container, which I assume is because of the crypto libraries in the fips image being used. However, Dominic and Takuro are the ones who created that container (I think) and so they would have a better idea of why that image is used instead of the devcontainer. I just didn't see a need to update that image.

ksayid (approved):

PR #111, opened 2024-04-01
update encfs to use cbl mariner

hgarvison:
Update ENCFS and SKR to use Azure Linux 3.0 as image base.

### DIFF
--- a/cmd/remotefs/azurefs.go
+++ b/cmd/remotefs/azurefs.go
@@ -35,6 +35,7 @@ var (
 	_azmountRun                    = azmountRun
 	_containerMountAzureFilesystem = containerMountAzureFilesystem
 	_cryptsetupOpen                = cryptsetupOpen
+	_cryptsetupOpenWithHeaderFile  = cryptsetupOpenWithHeaderFile
 	ioutilWriteFile                = os.WriteFile
 	osGetenv                       = os.Getenv
 	osMkdirAll                     = os.MkdirAll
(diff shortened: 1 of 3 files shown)


ksayid (approved):

docker/encfs/Dockerfile.encfs
```diff
@@ -7,9 +7,9 @@ RUN cd tools/get-snp-report && make && mv bin/get-snp-report /
 RUN CGO_ENABLED=0 GOOS=linux go build -o / ./cmd/azmount
 RUN CGO_ENABLED=0 GOOS=linux go build -o / ./cmd/remotefs
 
-FROM alpine:3.18.6
+FROM mcr.microsoft.com/cbl-mariner/base/core:2.0
 
-RUN apk update && apk upgrade --no-cache && apk add --no-cache cryptsetup fuse3 curl bash jq
+RUN tdnf update -y && tdnf upgrade -y && tdnf install -y cryptsetup fuse jq
```
BryceDFisher: add `&& tdnf clean all` at the end to remove cache files we don't need stored in the layer.

BryceDFisher (approved):

cmd/remotefs/azurefs.go
```diff
@@ -307,7 +333,10 @@ func containerMountAzureFilesystem(tempDir string, index int, fs AzureFilesystem
 	logrus.Debugf("Opening device at: %s", deviceNamePath)
 	err = _cryptsetupOpen(imageLocalFile, deviceName, keyFilePath)
 	if err != nil {
-		return errors.Wrapf(err, "luksOpen failed: %s", deviceName)
+		err = _cryptsetupOpenWithHeaderFile(imageLocalFile, deviceName, keyFilePath)
```
ksayid, line 347: Just wondering, if `cryptsetupOpen` fails, does the state change? Do we wanna do a `cryptsetup close` before doing another `cryptsetup open`?

ksayid: We could also add a log here, something like:
`logrus.Warnf("Initial cryptsetup open failed: %v. Attempting fallback approach with separate LUKS header...", err)`

cmd/remotefs/azurefs.go
```diff
@@ -88,11 +89,36 @@ func cryptsetupCommand(args []string) error {
...
+		return errors.Wrapf(err, "failed to redirect LUKS header into a separate file: %s", string(output))
+	}
+	logrus.Info("LUKS header file created: luksheader.img")
+
+	logrus.Info("Opening device with LUKS header file luksheader.img")
+	openArgs := []string{
+		// Open device with the key passed to luksFormat
+		"open", source, deviceName, "--key-file", keyFilePath,
+		// Use the header file
+		"--header=luksheader.img",
```
ksayid: should the header file be cleaned up after it's used?

ksayid (approved):

cmd/remotefs/azurefs.go
hgarvison: > Just wondering, if `cryptsetupOpen` fails, does the state change? Do we wanna do a `cryptsetup close` before doing another `cryptsetup open`?

No. It just fails to open the blob and leaves it alone.

ksayid (approved):
LGTM

PR #114, opened 2024-05-15
converting debug and example image to mariner

SethHollandsworth:
These two dockerfiles were still using images hosted on docker hub. This PR converts them to use Mariner as well as simplifies the interface to build them.

As it is, the context needed for `Dockerfile.debug` needs to be `./docker/skr` while `Dockerfile.example` uses the root of the repo as its context. `Dockerfile.debug` now uses the root as well to simplify

### DIFF
--- a/docker/skr/build-debug.sh
+++ b/docker/skr/build-debug.sh
@@ -8,20 +8,20 @@ pushd $(dirname $0)
 
 # This script builds the binaries and sets up the docker image
 
-mkdir -p bin
-pushd bin
+mkdir -p ../../bin
+pushd ../../bin
 CGO_ENABLED=0 GOOS=linux go build github.com/Microsoft/confidential-sidecar-containers/cmd/skr
 popd
 
 pushd ../../tools/get-snp-report
-make 
+make
 popd
 
-cp ../../tools/get-snp-report/bin/get-snp-report ./bin
-cp ../../tools/get-snp-report/bin/get-fake-snp-report ./bin
-cp ../../tools/get-snp-report/bin/verbose-report ./bin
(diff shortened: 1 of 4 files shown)


docker/skr/build-debug.sh
```diff
@@ -8,20 +8,20 @@ pushd $(dirname $0)
...
 
 pushd ../../tools/get-snp-report
-make 
+make
 popd
 
-cp ../../tools/get-snp-report/bin/get-snp-report ./bin
-cp ../../tools/get-snp-report/bin/get-fake-snp-report ./bin
-cp ../../tools/get-snp-report/bin/verbose-report ./bin
+cp ../../tools/get-snp-report/bin/get-snp-report ../../bin/
```
hgarvison, line 20: Can we do what we did for SKR and ENCFS and put steps like this in the Dockerfile?

SethHollandsworth: like make it a multi-stage build? we can/should do that, but I think it would be better as a separate PR

hgarvison (approved):

docker/skr/build-debug.sh
```diff
@@ -8,20 +8,20 @@ pushd $(dirname $0)
...
+cp ../../tools/get-snp-report/bin/get-fake-snp-report ../../bin/
+cp ../../tools/get-snp-report/bin/verbose-report ../../bin/
 
-docker build --tag skr -f Dockerfile.debug .
+docker build --tag skr -f Dockerfile.debug ../..
 
 # cleanup
 rm -rf bin
```
ksayid, line 27: should `bin` be changed to `../../bin`?

ksayid (approved):

stevendongatmsft (approved):
LGTM

PR #115, opened 2024-05-30
Add Attestation Test similar to Examples Repo

DomAyre:
With [confidential-aci-testing](https://github.com/microsoft/confidential-aci-testing)  we can now add tests similar to what we have in confidential-aci-examples with minimal code in this repo.

Added the attestation test as a proof of concept, can merge without the others and they can be added later

Todo:
- [x] Add a GitHub actions workflow
- [x] Make sure it works in fresh environment particularly in a codespaces testing view
- [x] ~Find an MCR python image which works~ For future PR's
- [x] Understand Analyse workflow failures and fix


.github/workflows/ci.yml
```diff
@@ -106,4 +110,31 @@ jobs:
         # Can't use `go test -v ./...` because some of the tests depend on SEV-SNP or Confidential ACI features
         go test -v ./... -tags skip_e2e,skip_snp_required
 
+  test-attestation:
```
hgarvison, line 112: Do you think each sidecar should be its own job or should we have tests for each sidecar under this job to avoid duplicating the steps of installing the testing package?

tests/attestation/attestation.bicepparam
```diff
@@ -0,0 +1,12 @@
+using './attestation.bicep'
+
+// Image info
+param registry='confidentialsidecars.azurecr.io'
+param tag=''
+
+// Deployment info
+param location='westeurope'
```
hgarvison, line 8: Do you think we should do a matrix env var for location in the ci step and fill this value in? It would mimic the current examples repo and allow us to test in the different regions in this repo instead of the examples (if we're trying to remove the examples repo).

tests/attestation/attestation.bicep
```diff
@@ -0,0 +1,105 @@
...
+            {
+              protocol: 'TCP'
+              port: 8080
+            }
+          ]
+          command: ['app', '-socket-address', '/mnt/uds/sock']
+          environmentVariables: [
+            {
+              name: 'AZURE_ATTESTATION_ENDPOINT'
+              value: 'caciexamples.eus.attest.azure.net'
```
hgarvison: Should we add something like this to the env file/secrets list like we did in the examples repo?

tests/attestation/attestation.bicepparam
DomAyre: I think the purpose here less about testing ACI which is (partially) the use case of the examples repo, and more about testing the sidecars themselves, so I don't think it's too important to test different regions, except for maybe having `eastus2euap` since that often will get new features first which might be good to know about (e.g. the new kernel versions)

tests/attestation/attestation.bicep
DomAyre: Yeah I think we should at least make one for sidecars specifically, but also making it configurable wouldn't hurt

tests/attestation/attestation.bicepparam
DomAyre: What I've settled on is having a separate workflow file which runs regularly and on demand to test the images in canary, I don't think failures in canary should block PRs

tests/attestation/attestation.bicep
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/115/commits/7f66a1a2355d5e06b8d2d9459cbfb099ecd930af

.github/workflows/ci.yml
DomAyre: I think the benefit of having each sidecars test running independently is worth the bit of duplicated setup?

tests/attestation/primary.Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM python:latest
```
SethHollandsworth, line 1: the image I've been using is: mcr.microsoft.com/cbl-mariner/base/python:3.9 . worth a try to see if it works imo

tests/attestation/test.py
```diff
@@ -0,0 +1,255 @@
...
+        launch_tcb,
+        signature,
+    ) = struct.unpack_from(f"<{SNP_REPORT_STRUCTURE}", attestation_bytes, 0)
+
+    # Validate report was generated based on the provided report data
+    # (i.e. This is the report we requested)
+    seen_report_data = report_data.rstrip(b"\x00").decode()
+    print(f"Checking that report data in the attestation: {seen_report_data}")
+    print(f"Matches the report data we provided: {expected_report_data}")
+    assert seen_report_data == expected_report_data
```
SethHollandsworth: Logging would be cleaner if there was a failure message along with the assert rather than print statements before and after the assert

e.g.
`assert seen_report_data == expected_report_data, "Expected report data does not match what was seen"`

These could also be f-strings if you want to have both pieces of data printed out

tests/attestation/primary.Dockerfile
DomAyre: I tried it initially and unfortunately it didn't work out of the box, but I'm sure if can be made to work with some adjustments

tests/attestation/test.py
DomAyre: I agree except I think it's nice to see the logs even in the success case, which you don't get with the following message

tests/attestation/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    image: $REGISTRY/attestation/sidecar:$TAG
+    build:
+      context: ../../
+      dockerfile: docker/attestation-container/Dockerfile.run
+    command: ["app", "-socket-address", "/mnt/uds/sock"]
+    network_mode: "host"
+    ports:
+      - "8080:8080"
+    environment:
+      - AZURE_ATTESTATION_ENDPOINT=cacisidecars.eus.attest.azure.net
```
hgarvison: Do we want this as a secret/env variable as well?

DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/115/commits/05922b3fb3d0743ec4af3730a125f5e9f68461e4

tests/attestation/primary.Dockerfile
DomAyre: Couldn't easily figure this out so I suggest we merge as is for now

hgarvison (approved):

PR #116, opened 2024-06-03
remove trivy scan exit code

hgarvison:
Remove Trivy Security Scan exit code so that the scan results get uploaded.


SethHollandsworth (approved):
lgtm

PR #117, opened 2024-06-04
Add an SKR test similar to Examples repo

DomAyre:


.github/workflows/ci.yml
```diff
@@ -136,4 +139,53 @@ jobs:
...
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.MANAGED_ID_CLIENT_ID }}
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: ${{ vars.SUBSCRIPTION }}
+
+      # Deployment is separated from testing to avoid the AZ CLI login timeout
+      - name: Deploy the SKR container group
+        env:
+          ID: ${{ env.BASE_ID }}-skr
```
hgarvison: Nit: maybe switch the order to skr-base_id?

deployment/deploy_managed_hsm.sh
```diff
@@ -0,0 +1,50 @@
...
+# Licensed under the MIT License.
+
+set -e
+
+echo "There are three steps to deploying the managed HSM"
+echo " - Deploy the HSM in azure"
+echo " - Create 3 keys to encrypt the security domain"
+echo " - Download the security domain and activate the HSM"
+
+output=$(az deployment group create \
```
hgarvison, line 13: Do we have a cleanup step for this somewhere? I think it would error if you try to create the hsm each run and it already exists...

tests/skr/proxy.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import base64
+import json
+import subprocess
+from flask import Flask, request, Response
+import requests
+
+app = Flask(__name__)
```
hgarvison, line 10: Why the switch to Flask from the http lib? We used Flask with the Hello World ACI demo but ended up having problems with Flask in conf AKS, so we switched to NGINX for the Hello World AKS demo. I don't think it matters for this test, but curious as to why the switch from the ACI Examples repo tests.

tests/skr/proxy.Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM fullstorydev/grpcurl:v1.9.1 as grpcurl-builder
```
SethHollandsworth: [this file](https://github.com/microsoft/confidential-sidecar-containers/blob/main/docker/skr/Dockerfile.example) has a way of downloading grpcurl without needing a second image via downloading the package from github and installing. 

This limits our usage on non-MCR registries

tests/skr/proxy.Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+FROM python:latest
```
SethHollandsworth, line 2: We should use an image like mcr.microsoft.com/cbl-mariner/base/python:3.9. I tried it with a minimal Flask example, and it worked. As well for using grpc, [this PR](https://github.com/microsoft/confidential-sidecar-containers/pull/114/files) has it on a mariner image

tests/skr/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
+services:
+  proxy:
+    image: $REGISTRY/skr/proxy:$TAG
+    network_mode: "host"
+    build:
+      context: .
+      dockerfile: proxy.Dockerfile
+  http_sidecar:
+    image: $REGISTRY/skr/http_sidecar:$TAG
```
SethHollandsworth: If the http and grpc sidecars are the same image, using the same image name and only pushing from one of these build statements would be simpler in my opinion

tests/skr/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
+version: '3'
+services:
+  proxy:
+    image: $REGISTRY/skr/proxy:$TAG
```
hgarvison, line 4: Will cleanup still take care of these images with the '/' included?

.github/workflows/ci.yml
```diff
@@ -136,4 +139,53 @@ jobs:
           source cacitesting.env
           python -m unittest discover -s tests/attestation
 
+  test-skr:
```
hgarvison, line 144: We should add a cleanup job to this file so that things get cleaned up after the run

tests/skr/test.py
```diff
@@ -0,0 +1,282 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+
+        cls.target_dir = os.path.realpath(os.path.dirname(__file__))
+        cls.id = os.getenv("ID", str(uuid.uuid4()))
+        cls.tag = os.getenv("TAG") or cls.id
+
+        cls.attestation_endpoint = "confidentialsidecars.weu.attest.azure.net"
+        cls.hsm_endpoint = "cacisidecars.managedhsm.azure.net"
```
SethHollandsworth: I believe these 2 should be env vars/ secrets. and their names in the corresponding bicep files

tests/skr/proxy.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
SethHollandsworth, line 1: is the purpose of this container that grpcurl isn't on the test runner machine? It seems like indirection to wrap the grpc server with an http server if we could call the grpc directly from test.py

deployment/deploy_managed_hsm.sh
DomAyre: The managed HSM is not deployed per run, since it's non-trivial to setup, it's done once and is long lived. So this is as much about documentation of how to spin up one than anything else.

tests/skr/proxy.py
DomAyre: No special reason other than I found Flask a little more concise to express the endpoints

tests/skr/docker-compose.yml
DomAyre: You're right I forgot the cleanup step only cleaned up images where the tag matches the branch, I'll update it to delete all old images

.github/workflows/ci.yml
DomAyre: The c_aci_testing package handles cleanup for us, when `target_run_ctx` exits it cleans up deployed resources

tests/skr/proxy.Dockerfile
DomAyre: That's a fair reason I opted for the image based approach to make the docker file simpler, but I hadn't considered the requirement to prefer not using MCR images (although I guess wget'ing some source code is probably just as bad)

tests/skr/docker-compose.yml
DomAyre: You're right, the only reason to split them here would be to make running them locally easier but since they rely on being in ACI I guess that's not a likely use case

tests/skr/test.py
DomAyre: Sure I can migrate them over

tests/skr/proxy.py
DomAyre: It's mostly to test/demonstrate them as sidecars to some primary container since users would more likely use that pattern

.github/workflows/ci.yml
DomAyre: Sounds good

tests/skr/docker-compose.yml
DomAyre: Actually, maybe it's worth waiting until all testing has moved over to being done here? That way we don't break the currently running tests in examples? I think that's worth having some slightly longer living images for now

.github/workflows/ci.yml
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/84622e314ca2f451e049645f2611e42652b528e2

tests/skr/docker-compose.yml
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/176edbded9fee3240539a4b70607c9c8278cd6aa

tests/skr/proxy.Dockerfile
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/9669a1ed456058e488428614ffd6f43bd7343e61

tests/skr/test.py
DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/845fa0e5951e270c313ee3c7594f8370b913ee32

tests/skr/proxy.Dockerfile
DomAyre: I tried in this commit
https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/81bdd9e82a7b27d1ccabf1847895c63f21a13ab1

Is there any obvious mistake I'm making? Running the image locally works fine, but in ACI it almost immediately enters the terminated state
https://github.com/microsoft/confidential-sidecar-containers/actions/runs/9447436790/job/26019227329?pr=117

tests/skr/docker-compose.yml
hgarvison: That works. Since the cleanup removes them all, they'll get removed when you that later.

tests/skr/proxy.Dockerfile
hgarvison: There's a long standing issue with ACI where we can't exec into the mariner images and it enters the terminated state. This issue is why I haven't merged the conversion of ENCFS to Mariner. They are aware of it, but no ETA on when the fix will come out.

.github/workflows/ci.yml
hgarvison: We should probably add some info on this in the README or even a new README in the workflows dir

tests/skr/proxy.py
hgarvison: As far as testing goes, we could utilize flask to actually make an html webpage where we output debug info so that we don't have to exec into the container (adding the --debug-mode flag to the policy) each time it fails.

deployment/deploy_managed_hsm.sh
hgarvison: We should probably add a README to the workflows dir to explain things like this for future development and other developers not in this PR.

DomAyre: Sounds good, would you like it as part of this review or I can write something up with the final PR for encfs?

tests/skr/proxy.py
DomAyre: We can, `c_aci_testing` also helps in this use case by printing out all container logs on cleanup

tests/skr/proxy.Dockerfile
DomAyre: That's interesting, I suspect this could be related but I don't think it's the same as I don't exec into the containers. I think while it's still not fully supported I would opt for sticking to the working base image

deployment/deploy_managed_hsm.sh
hgarvison: I think it should be fine to include with the ENCFS PR

tests/skr/proxy.Dockerfile
hgarvison: That's fine by me.

hgarvison (approved):

PR #118, opened 2024-06-10
Move attestation endpoint into an env variable

DomAyre:
Spun out from #117 since it's not really related to SKR

### DIFF
--- a/tests/attestation/attestation.bicepparam
+++ b/tests/attestation/attestation.bicepparam
@@ -10,4 +10,4 @@ param ccePolicies={
   attestation: ''
 }
 param managedIDName='cacisidecars'
-param attestationEndpoint='https://confidentialsidecars.weu.attest.azure.net'
+param attestationEndpoint=''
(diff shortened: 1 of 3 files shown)


tests/attestation/attestation.bicepparam
```diff
@@ -10,4 +10,4 @@ param ccePolicies={
   attestation: ''
 }
 param managedIDName='cacisidecars'
-param attestationEndpoint='https://confidentialsidecars.weu.attest.azure.net'
+param attestationEndpoint=''
```
hgarvison, line 13: Do we need to have the param already existing in this file or can we use the test.py file to insert the whole line instead of having a placeholder? Not sure which is better tbh, but this isn't a dealbreaker for me so merge if this way is better.

hgarvison (approved):

tests/attestation/attestation.bicepparam
DomAyre: It could be committed, but it would be set at runtime either way based on the env variable

PR #119, opened 2024-06-11
Fix cleanup workflow

DomAyre:
It was missing permissions and an incorrect RG name


DomAyre:
Cleanup workflow manually run here:
https://github.com/microsoft/confidential-sidecar-containers/actions/runs/9463424499

PR #121, opened 2024-06-12
Add an EncFS Test similar to the Examples Repo

DomAyre:
Adds an encrypted filesystem test equivalent to what we have in the examples repo

Todo:
- [x] Add testing for read/write
- [x] Add testing for block fs's
- [x] Update cleanup job


scripts/install-c-aci-testing.sh
```diff
@@ -2,5 +2,5 @@
 
 latest=$(gh release list -R microsoft/confidential-aci-testing -L 1 --json tagName --jq '.[0].tagName')
 gh release download $latest -R microsoft/confidential-aci-testing
```
ksayid, line 6: nits: 
* error handling could be added to the `gh release list` / `download` in case of network issues
* `set -e` could be added to the shell scripts (and the run blocks of the GitHub Actions workflows)

scripts/setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+sudo apt-get update -y
```
hgarvison: Should probably do an upgrade too and I don't believe update requires -y, just upgrade:

scripts/install-c-aci-testing.sh
```diff
@@ -2,5 +2,5 @@
...
-pip install c-aci-testing*.tar.gz
+python -m pip install c-aci-testing*.tar.gz
```
hgarvison, line 7: Rather than using the setup script to call this script, can we just merge this script into the setup script? Seems part of setup anyway

.github/workflows/ci.yml
```diff
@@ -192,4 +192,38 @@ jobs:
...
+        run: sudo usermod -aG docker $USER
+
+      - name: Log into Azure
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.MANAGED_ID_CLIENT_ID }}
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: ${{ vars.SUBSCRIPTION }}
+
+      - name: Test EncFs
```
hgarvison: Nit: consistency with the capitalization of ENCFS. I usually do all caps to match SKR despite the capitalization not matching the number of words.

.github/workflows/ci.yml
```diff
@@ -192,4 +192,38 @@ jobs:
...
+    name: Test EncFS
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install c_aci_testing package
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: ./scripts/install-c-aci-testing.sh
```
hgarvison, line 214: If you merge setup and this script, can change this to setup.sh

deployment/blobStorage.bicep
```diff
@@ -0,0 +1,66 @@
...
+  properties: {
+    publicAccess: 'None'
+  }
+}
+
+
+resource managedIdentityRoleAssignment 'Microsoft.Authorization/roleAssignments@2022-04-01' = {
+    name: guid(storageAccount.id, 'MI', 'Storage Blob Data Contributor')
+    properties: {
+      roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')  // Storage Blob Data Contributor Role ID
```
hgarvison, line 46: Should the role ID be a secret?

deployment/blobStorage.bicep
```diff
@@ -0,0 +1,66 @@
...
+    }
+    scope: resourceGroup()
+  }
+
+
+resource roleAssignments 'Microsoft.Authorization/roleAssignments@2022-04-01' = [
+  for id in otherIds: {
+    name: guid(storageAccount.id, id, 'Storage Blob Data Contributor 2')
+    properties: {
+      roleDefinitionId: subscriptionResourceId('Microsoft.Authorization/roleDefinitions', 'ba92f5b4-2d11-453d-a403-e96b0029c9fe')  // Storage Blob Data Contributor Role ID
```
hgarvison, line 57: Secret?

tests/encfs/primary.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:latest
```
ksayid, line 1: nit: should this be the python image from MCR?

tests/encfs/primary.Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+WORKDIR /usr/src/app
+
+RUN pip install flask
+
+COPY primary.py .
+
+CMD ["python3", "primary.py"]
```
hgarvison, line 8: Nit: can this be just python like it is in the examples repo?

tests/encfs/encfs.py
```diff
@@ -0,0 +1,103 @@
...
+                ]
+            ),
+            shell=True,
+        )
+
+    def __init__(self, key_path, image_path):
+        self.key_path = key_path
+        self.image_path = image_path
+        with open(image_path, "wb") as f:
+            f.seek(64 * 1024 * 1024 - 1)
```
hgarvison, line 39: Instead of doing the math, can we make this a constant?

.github/workflows/ci.yml
```diff
@@ -192,4 +192,38 @@ jobs:
           source cacitesting.env
           python -m unittest discover -s tests/skr
 
+  test-encfs:
```
ksayid, line 204: If the tests depend on the `build` job, then I think the [`needs` ](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) attribute should be used to define this dependency.

deployment/blobStorage.bicep
```diff
@@ -0,0 +1,66 @@
...
+resource storageAccount 'Microsoft.Storage/storageAccounts@2021-04-01' = {
+  name: '${name}storage'
+  location: location
+  kind: 'StorageV2'
+  sku: {
+    name: sku
+  }
+  properties: {
+    accessTier: 'Hot'
+    allowBlobPublicAccess: true
```
ksayid: do we need to have public blob access?

.github/workflows/cleanup.yml
```diff
@@ -67,3 +64,66 @@ jobs:
...
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: ${{ vars.SUBSCRIPTION }}
+
+      - name: Delete HSM Keys
+        run: |
+          az keyvault key list --hsm-name ${{ vars.HSM_NAME }} \
+            --query "[].kid" -o tsv | \
+            xargs -I{} az keyvault key delete --id {}
+
+  cleanup_registry:
```
ksayid, line 50: nit: cleanup registry is a job that appears twice (L13 and L85), which might be a little confusing. Giving each a more unique name that clarifies the purpose of each would be helpful.

.github/workflows/cleanup.yml
```diff
@@ -67,3 +64,66 @@ jobs:
...
+      - name: Log into Azure
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.MANAGED_ID_CLIENT_ID }}
+          tenant-id: ${{ secrets.MANAGED_ID_TENANT_ID }}
+          subscription-id: ${{ vars.SUBSCRIPTION }}
+
+      - name: Delete Blobs
+        run: |
+          az storage blob list \
```
ksayid, line 86: nit: for resource deletion in general, maybe more logging / error handling could be added?

tests/encfs/test.py
```diff
@@ -0,0 +1,172 @@
...
+    def tearDownClass(cls):
+        # Cleans up the ACI instance
+        cls.aci_context.__exit__(None, None, None)
+
+    def test_read_rw_encfs(self):
+
+        response = requests.get(
+            f"http://{self.encfs_ip}:8000/read_file?path={self.blobs[0][0]}/file.txt",
+        )
+        assert response.status_code == 200
```
ksayid: nit: adding messages to assertions would be helpful to provide context on failures.
logging would also be beneficial throughout the test flow.

tests/encfs/test.py
```diff
@@ -0,0 +1,172 @@
...
+
+        azure_args = {
+            "subscription": os.getenv("SUBSCRIPTION"),
+            "resource_group": os.getenv("RESOURCE_GROUP")
+        }
+
+        image_args = {
+            "target": target_dir,
+            "registry": os.environ["REGISTRY"],
+            "repository": os.getenv("REPOSITORY"),
```
ksayid, line 59: nit: env vars could be validated before using them. if this isn't a risk, I think it's fine though.

tests/encfs/primary.py
```diff
@@ -0,0 +1,30 @@
...
+@app.route('/write_file', methods=['POST'])
+def write_file():
+
+    file_path = os.path.join(ENCFS_MOUNT, request.args.get("path"))
+    try:
+        with open(file_path, "w") as f:
+            f.write(request.data.decode())
+        return f'{request.args.get("path")} written to', 200
+    except OSError as e:
+        return e.strerror, 400
```
ksayid: nit: maybe status code 500 would be more appropriate?

tests/encfs/primary.py
```diff
@@ -0,0 +1,30 @@
...
+ENCFS_MOUNT=os.environ["ENCFS_MOUNT"]
+
+@app.route('/read_file', methods=['GET'])
+def read_file():
+
+    file_path = os.path.join(ENCFS_MOUNT, request.args.get("path"))
+    if not os.path.exists(file_path):
+        return {"error": "File not found"}, 404
+
+    with open(file_path) as f:
```
ksayid, line 15: like the case for `write_file`, should there be a try-except here for the `with` block?

tests/encfs/encfs.py
```diff
@@ -0,0 +1,103 @@
...
+            f.write(b"\0")
+
+    def __enter__(self):
```
ksayid, line 42: nit: maybe more logging and error handling?

tests/encfs/encfs.py
```diff
@@ -0,0 +1,103 @@
...
+        subprocess.check_call(
+            f"sudo mount -t ext4 {self.DEVICE_NAME_PATH} {self._dir.name} -o loop",
+            shell=True,
+        )
+        return self._dir.name
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        subprocess.check_call(f"sudo umount {self._dir.name}", shell=True)
+        self._run_command("luksClose", self.DEVICE_NAME)
+        self._dir.cleanup()
```
ksayid: nit: maybe try-finally here to ensure `cleanup()` is always run

.github/workflows/ci.yml
DomAyre: You're right, but they don't depend on that job c_aci_testing handles everything so I think we're fine here

.github/workflows/ci.yml
DomAyre: setup.sh is more for local development, for ci, we only need the testing package so I think it makes sense to keep them separate

.github/workflows/ci.yml
DomAyre: Done in 6848b91e252ea9fa4afc4bff0f529d380b72ff7c

.github/workflows/cleanup.yml
DomAyre: Up to you, I think it's a best effort to stop things like storage accounts getting too full, if it fails as a one off I don't think it matters too much, and if it started failing consistently we would see the job history and take more action

.github/workflows/cleanup.yml
DomAyre: Yeah so the first job is going to be superceded by these new tests, instead of uploading images to a registry which the examples repo pulls to test, testing will happen here. I was imagining a separate cleanup PR but since this one completes the replacement I'll remove the old stuff now

DomAyre: Done in 34970d2f68d49a242b0c011d3531baa6e9b3987f

deployment/blobStorage.bicep
DomAyre: I believe we do since the github action runners and the ACI's need access, access is still role based. Having said that, I see the default value for this is true so this line isn't actually necessary

DomAyre: Removed in 1c39ab12f3d83c7664549684dafc55a31692af75

deployment/blobStorage.bicep
DomAyre: I don't believe so, Storage Blob Data Contributor is a built in azure role and it's ID is public knowledge, the only reason to make it a repo variable is we need to change it, but I don't see that happening

deployment/blobStorage.bicep
DomAyre: See my other comment

scripts/install-c-aci-testing.sh
DomAyre: I think I've mentioned in another reply, but setting up the testing package is the only thing we need in CI, but for local development we need more which is what setup.sh is for, so I think keeping them separate makes sense, also it's handy if I release a new version of the testing package to have a target just for that

scripts/install-c-aci-testing.sh
DomAyre: I think `set -e` alone is a good bang for buck solution regarding error handling given the likelyhood of failure

DomAyre: Done in 02cd7c47585c8d2f7ae6777b9ae68973dba5ba05

scripts/setup.sh
DomAyre: You're right, done in 27a1642d84c3319d67756924bb44afbf099c2f03

tests/encfs/encfs.py
DomAyre: I've done some refactoring to make cleanup more robust here bbd4a68db95b1ccc05a7e13191eb93812348fbb9

tests/encfs/encfs.py
DomAyre: Done in bbd4a68db95b1ccc05a7e13191eb93812348fbb9

tests/encfs/encfs.py
DomAyre: Up to you, I think the math acts as documentation that its 64mb

tests/encfs/primary.Dockerfile
DomAyre: I think generally speaking it's better to specify 3 but up to you

tests/encfs/primary.Dockerfile
DomAyre: I tried swapping it out in my last [PR](https://github.com/microsoft/confidential-sidecar-containers/pull/117/commits/81bdd9e82a7b27d1ccabf1847895c63f21a13ab1) but it fails I think Heather mentioned some issue with the image. It immediately terminates for me

tests/encfs/primary.py
DomAyre: Up to you, its included in the write case because we test read only filesystems and therefore have tests expecting it to fail. If this fails flask will still return HTTP 500 so I don't personally think it's necessary

tests/encfs/primary.py
DomAyre: Probably, I used 400 to match the version in the examples repo, @hgarvison was there a particular reason to use 400 over 500?

tests/encfs/test.py
DomAyre: If you mean validated that they exist, I've used os.environ for registry for that reason so we get a keyerror if it's not set. For repository the testing package allows None which is why I use getenv() instead

tests/encfs/test.py
DomAyre: I've added response content printing on failure here: b62346c6741b17dda4da963c712aec04934ee211

The testing package already does quite a bit of logging, including printing the container logs so I hope that's sufficient?

.github/workflows/cleanup.yml
ksayid: Gotcha, makes sense. I think it'd be helpful to be able to easily tie the resources back to the test run for RCA if something like consistent failure were to happen. Is there already a way to do that currently? If not, logs or just adding a tag to the resource through the bicep file ([example](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/tag-resources-bicep#apply-values)) could work

tests/encfs/primary.Dockerfile
hgarvison: Doesn't make a difference to me. Sometimes specifying the version (or not specifying it) can cause errors in the container if it's not aliased properly. So as long as it works, it's fine as is.

tests/encfs/encfs.py
hgarvison: A constant with 64MB in the name would do the same, but I'll leave the decision up to you.

tests/encfs/primary.py
hgarvison: I think you did the initial error code definitions in the HTTP servers for the tests and if I made the tests, I just copied the formatting. So no special reasoning from me. Though looking at it now, I agree that 500 is probably better.

DomAyre: Haha I have the memory of a sieve, great if we're all happy with 500 I'll change it

DomAyre: https://github.com/microsoft/confidential-sidecar-containers/pull/121/commits/c017afe7ceff342efe2c9c3fabb844a5ea3fab01

.github/workflows/cleanup.yml
DomAyre: Yeah I can add a tag that a given resource came from this repo if you'd like? They all go into the same `confidential-sidecars` resource group so I think that would make tracing easy enough but happy to take input on that

hgarvison (approved):

.github/workflows/cleanup.yml
ksayid: Gotcha, I was thinking of a direct tie between the GitHub run number/id and the resources created. Like you mentioned, we would already know the that they came from this repo since they all go into the same resource group. But I think having a way to relate resource to run would be convenient. For example, suppose you find a resource in the resource group (so it wasn't cleaned up). Having tags on the resource specifying the run numbers/ids would make it easy to see what runs to lookup to root cause the issue.

SethHollandsworth (approved):

.github/workflows/cleanup.yml
DomAyre: Makes sense, for that use case, the container groups are named based on the run id and run number so that can be used to relate back to the run

PR #122, opened 2024-06-13
remove remaining aasp

stevendongatmsft:


pkg/grpc/grpcserver/server.go
```diff
@@ -49,7 +49,7 @@ type AzureInformation struct {
 }
 
 const (
-	AASP = "aasp"
+	ATTESTATIONAGENT = "skr"
```
hgarvison: I would put a '_' in between the words to make it clearer.

tests/skr/proxy.py
```diff
@@ -51,7 +51,7 @@ def grpc_unwrap_key():
                     "dc": {
                         "Parameters": {
                             "attestation-agent": [
-                                base64.b64encode("aasp".encode()).decode()
+                                base64.b64encode("skr".encode()).decode()
```
hgarvison, line 54: You need to fix unwrap.sh as well

hgarvison (approved):

PR #123, opened 2024-06-14
Fix canary test

DomAyre:
Couple of small errors:

- Imports are misinterpreted when running in unittest outside the test directory
- __init__.py is missing from skr dir so test isn't found


DomAyre:
While the workflow is fixed, canary region appears to currently be broken
https://github.com/microsoft/confidential-sidecar-containers/actions/runs/9514183107

Which is backed up by the history in examples
https://github.com/microsoft/confidential-aci-examples/actions/workflows/status_eastus2euap.yml

PR #127, opened 2024-06-20
update maa test base image to mariner python

hgarvison:
Update maa test base image to mariner python and update curl command.


ksayid (approved):
LGTM

PR #134, opened 2024-07-17
Use log level debug in encfs

DomAyre:


tests/encfs/encfs.bicep
```diff
@@ -91,6 +91,10 @@ resource containerGroup 'Microsoft.ContainerInstance/containerGroups@2023-05-01'
               name: 'EncfsSideCarArgs'
               value: sidecarArgsB64
             }
+            {
+              name: 'LogLevel'
```
hgarvison, line 95: We can do the same thing with SKR

hgarvison (approved):

PR #140, opened 2024-09-17
updating skr pipeline test to have debug level logging

hgarvison:
updating skr pipeline test to have debug level logging

### DIFF
--- a/tests/skr/skr.bicep
+++ b/tests/skr/skr.bicep
@@ -59,6 +59,12 @@ resource containerGroup 'Microsoft.ContainerInstance/containerGroups@2023-05-01'
         name: 'http-sidecar'
         properties: {
           image: '${registry}/skr/sidecar:${empty(tag) ? 'latest': tag}'
+          environmentVariables: [
+            {
+              name: 'LogLevel'
+              value: 'debug'
+            }
+          ]
           ports: [
             {
               protocol: 'TCP'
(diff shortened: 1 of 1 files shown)


KenGordon:
Heather, why do you want this? I am a bit worried that it may encourage users who do not pay attention to do the wrong thing.
Could you have an extra workflow or test with an obvious name (reckless_with_secrets_test or something)?

hgarvison:
@KenGordon I saw that Dominic had made a PR to add debug logging to ENCFS (I didn't realize at the time that he was using it to debug something specific) and I figured it could be useful in SKR too. The debug level logging should only be able to be accessed by team members who have access to the deployment itself, so I don't see a security issue, and it could potentially help someone debug future edits to SKR. However, I'm open to leaving it excluded and having people include it in a case-by-case basis.

KenGordon:
> @KenGordon I saw that Dominic had made a PR to add debug logging to ENCFS (I didn't realize at the time that he was using it to debug something specific) and I figured it could be useful in SKR too. The debug level logging should only be able to be accessed by team members who have access to the deployment itself, so I don't see a security issue, and it could potentially help someone debug future edits to SKR. However, I'm open to leaving it excluded and having people include it in a case-by-case basis.

I think the risk is people copy and paste the test bicep files as they are more or less what they want without being 100% certain they understand every line. I don't have a terribly strong position, I guess we should consider backing out Dom's change if we don't want this similar one.

PR #141, opened 2024-09-25
update skr to allow for ecdsa key and fix null reference error

hgarvison:
update skr to allow for ecdsa key and fix null reference error with invalid key type

### DIFF
--- a/pkg/skr/skr.go
+++ b/pkg/skr/skr.go
@@ -5,6 +5,7 @@ package skr
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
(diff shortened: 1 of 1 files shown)


ksayid (approved):

pkg/skr/skr.go
```diff
@@ -138,7 +140,22 @@ func SecureKeyRelease(identity common.Identity, certState attest.CertState, SKRK
...
+
+		jwKey := jwk.NewECDSAPrivateKey()
+		err = jwKey.FromRaw(privateEcdsaKey)
+		if err != nil {
+			return nil, errors.Wrapf(err, "could not encode ECDSA key as JWK")
+		}
+		return jwKey, nil
 	} else {
-		return nil, errors.Wrapf(err, "released key type not supported")
+		errString := fmt.Sprintf("released key type %v not supported", kty)
```
SethHollandsworth: nit: why not Wrapf like the one above?

ksayid: I guess there isn't an `err` to wrap like the other case. maybe `errors.Errorf` could be used instead here?

SethHollandsworth: good point, I think Errorf would be a little cleaner

hgarvison: I was following the previous standard laid out in a bunch of the other files where we create new errors.

hgarvison: Also, the wrapf was causing a null reference error that Mainak's team noticed because there was no error to wrap.

ksayid (approved):

SethHollandsworth (approved):
lgtm

PR #150, opened 2024-11-20
update skr maa token test Dockerfile to base64url encode MAA request

hgarvison:
update skr maa token test Dockerfile to base64url encode MAA request

### DIFF
--- a/docker/skr/Dockerfile.maa_test
+++ b/docker/skr/Dockerfile.maa_test
@@ -1,8 +1,8 @@
 FROM mcr.microsoft.com/cbl-mariner/base/python:3.9
 USER root
-RUN tdnf update -y && tdnf upgrade -y && tdnf install curl && tdnf clean all
+RUN tdnf update -y && tdnf upgrade -y && tdnf install curl coreutils && tdnf clean all
 RUN pip install jwcrypto
 # create a key for attaching to our request
 RUN python3 -c "from jwcrypto import jwk; print(jwk.JWK.generate(kty='RSA', size=2048, alg='RSA256').export_private())" > rsa.jwk
 # send request to get token from MAA via SKR container
-CMD sh -c 'until curl --fail --silent http://localhost:8080/status; do sleep 5; done; curl --fail-with-body -XPOST http://localhost:8080/attest/maa -d "{\"runtime_data\":\"$(cat rsa.jwk | base64 -w 0)\", \"maa_endpoint\": \"sharedeus2.eus2.test.attest.azure.net\"}"'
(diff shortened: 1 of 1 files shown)


ksayid (approved):

ksayid (approved):

KenGordon:
Can you double check that the code in the sidecar that consumes this is not expecting base64 butbase64url eg https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/internal/httpginendpoints/httpginendpoints.go#L137C82-L138C1 I think the actuial code that needs to match MAA is here https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/maa.go#L62 and particularly this line https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/maa.go#L119 for the customer in question I think they could set this flag to true https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/info.go#L22 and build a copy of the container, run it with a debug policy and watch all the steps of the construction of the request by looking at these logged artifacts.

hgarvison:
> Can you double check that the code in the sidecar that consumes this is not expecting base64 butbase64url eg https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/internal/httpginendpoints/httpginendpoints.go#L137C82-L138C1 I think the actuial code that needs to match MAA is here
> 
> https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/maa.go#L62
> 
> and particularly this line
> https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/maa.go#L119
> 
> for the customer in question I think they could set this flag to true
> https://github.com/microsoft/confidential-sidecar-containers/blob/8b7f6b8b49d0e50dde57f450acfc86775879b862/pkg/common/info.go#L22
> 
> and build a copy of the container, run it with a debug policy and watch all the steps of the construction of the request by looking at these logged artifacts.

Should this be the same for the RawAttest Runtime Data?

hgarvison:
@KenGordon the MAA docs are lacking specification as to whether base64url is required in all requests or not, but I have yet to see any docs that say base64 is required, versus several that specify base64url. I'm not sure who to talk to from MAA to confirm whether they should all be base64url or not.

PR #166, opened 2025-01-27
Fix importkey README example, and give more useful error messages on config parse fail

micromaomao:
README contains an example which doesn't work because importkey requires claims to be an array of array

### DIFF
--- a/tools/importkey/README.MD
+++ b/tools/importkey/README.MD
@@ -27,18 +27,20 @@ When importing a key to the key vault, a release policy is coupled with the key.
         "label": "Model Encryption Key"
     },
     "claims":[
-        {
-            "claim":  "x-ms-sevsnpvm-hostdata",
-            "equals": "99e1abd344fc0989288c44fde8de3b2a248b1b95814df8955d0c305a7db46680"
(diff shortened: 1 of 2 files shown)


tools/importkey/README.MD
```diff
@@ -27,18 +27,20 @@ When importing a key to the key vault, a release policy is coupled with the key.
...
-        },
-        {
-            "claim": "x-ms-compliance-status",
-            "equals": "azure-compliant-uvm"
-        },
-        {
-            "claim": "x-ms-sevsnpvm-is-debuggable",
-            "equals": "false"
-        }
+        [
```
hgarvison, line 30: Do we want this included? Seems to be an indentation change

micromaomao: The indentation change is because this is now inside an array. Alternatively if you don't want that I can not indent it

micromaomao: Sorry @hgarvison I should have made my PR clearer earlier. The example in README didn't work because importkey requires claims to be an array of array

hgarvison (approved):

PR #183, opened 2025-03-28
Update readme wrt virtual nodes.

KenGordon:
Add link to https://github.com/Azure/azure-cli-extensions/blob/main/src/confcom/azext_confcom/README.md#aks-virtual-node


KenGordon:
Too messy, will replace with single commit PR.

PR #185, opened 2025-03-28
Test_CertFetcher: use a newer report

micromaomao:
Also makes expected error check more robust to changes


pkg/attest/platform_cert_fetcher_test.go
```diff
@@ -81,9 +82,9 @@ func Test_CertFetcher(t *testing.T) {
 				TEEType:      ValidTEEType,
 				APIVersion:   ValidAPIVersion,
 			},
-			chipID:          "0xdeadbeef",
+			chipID:          "deadbeef",
```
KenGordon, line 85: Why the change?

KenGordon: Because the chipids are hex but don't have an 0x. (offline conversation)

KenGordon (approved):

PR #189, opened 2025-04-21
update token endpoint per msal library update

hgarvison:
update token endpoint from https://login.microsoftonline.com/%s/oauth2/token to https://login.microsoftonline.com/%s per msal library update

Updated example: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/README.md?plain=1#L51


SethHollandsworth:
Is there a link to the MSAL update? that would be helpful for reference

SethHollandsworth (approved):
lgtm

PR #199, opened 2025-05-07
remove version from docker-compose.yml because it's obsolete

hgarvison:
remove version from docker-compose.yml because it's obsolete to remove warning from tests


SethHollandsworth (approved):
lgtm

PR #207, opened 2025-05-21
Use Azure CLI 2.72.0

DomAyre:
### Why 

We have been seeing CI failures on main such as https://github.com/microsoft/confidential-sidecar-containers/actions/runs/15150328244

The error is 
```
ERROR: Can't get attribute 'NormalizedResponse' on <module 'msal.throttled_http_client' from '/opt/az/lib/python3.11/site-packages/msal/throttled_http_client.py'>
```

Based on the issue https://github.com/Azure/azure-cli/issues/31419 it looks like it is fixed in later versions so bumping here


KenGordon (approved):
Thank you.

KenGordon:
Thanks.

PR #208, opened 2025-05-21
update c-aci-testing package installation to install latest version

hgarvison:
update c-aci-testing package installation to install latest version from repo
remove attestation container


tests/attestation/test.py
```diff
@@ -34,7 +34,7 @@ def test_attestation(self):
 
         aci_param_set(
             target_path=target_dir,
-            parameters=[f'attestationEndpoint=\'https://{os.environ["ATTESTATION_ENDPOINT"]}\''],
+            parameters={"attestationEndpoint": f'\'https://{os.environ["ATTESTATION_ENDPOINT"]}\''},
```
ksayid: hmm so, attestationEndpoint needs to look like `"'https://foo'"`. are the single quotes redundant?

ksayid (approved):

tests/attestation/test.py
hgarvison: Based on this error, I don't think we're having trouble reaching this endpoint: https://github.com/microsoft/confidential-sidecar-containers/actions/runs/15419869459/job/43391813549?pr=208#step:7:811

ksayid: Gotcha but does it work without using both double and single quote?

hgarvison: Didn't work

ksayid: I guess specifically I meant something like this, which only has one quote
`f"https://{os.environ['ATTESTATION_ENDPOINT']}"`

ksayid: this is more a nit though, the tests are failing bc of a different issue I think

ksayid (approved):

Issue #213, opened 2025-06-03
CommittedSvn, CommittedVersion, and LaunchSvn are not serialized in SerializeReport


johannww:
The method SerializeReport() does not serialize the described fields, causing an error when the signature is checked:

https://github.com/microsoft/confidential-sidecar-containers/blob/e3e00a1a216a8087a71183f97bae1a1fab9adde2/pkg/attest/snp_attestation_report.go#L107

Possible solution:

```go
	for index, b := range chipIDByteArray {
		report[index+416] = b
	}

	binary.LittleEndian.PutUint64(report[480:488], r.CommittedSvn)
	binary.LittleEndian.PutUint64(report[488:496], r.CommittedVersion)
	binary.LittleEndian.PutUint64(report[496:504], r.LaunchSvn)

	reserved3ByteArray, err := hex.DecodeString(r.Reserved3)
	if err != nil {
		return nil, errors.Wrapf(err, "decoding r.Reserved3 failed")
	}
	for index, b := range reserved3ByteArray {
		report[index+504] = b
	}

```

I have tested the modfications in an API  that returns the SNPAttestationReport marshalled json:

```go
	r.GET("/report", func(c *gin.Context) {
		if server.DeserializedReport == nil {
			var err error
			server.DeserializedReport, err = report.DeserializedReport(server.ReportBytes)
			if err != nil {
				c.JSON(http.StatusInternalServerError, gin.H{"error": "Failed to deserialize report"})
				return
			}
		}

		c.JSON(http.StatusOK, server.DeserializedReport)
	})
```

I will propose a PR.

PR #215, opened 2025-06-03
Serialize CommitedSvn, CommittedVersion, and LaunchSvn

hgarvison:
Fixes serialization of the CommitedSvn, CommittedVersion, and LaunchSvn in the SerializeReport method.

Related to https://github.com/microsoft/confidential-sidecar-containers/issues/213.

### DIFF
--- a/pkg/attest/snp_attestation_report.go
+++ b/pkg/attest/snp_attestation_report.go
@@ -208,12 +208,18 @@ func (r *SNPAttestationReport) SerializeReport() (report []uint8, err error) {
 		report[index+416] = b
 	}
 
+	// Do not use, only intended for testing
+	// cannot be expected to always work in the face of new hardware from AMD
+	binary.LittleEndian.PutUint64(report[480:488], r.CommittedSvn)
+	binary.LittleEndian.PutUint64(report[488:496], r.CommittedVersion)
+	binary.LittleEndian.PutUint64(report[496:504], r.LaunchSvn)
+
 	reserved3ByteArray, err := hex.DecodeString(r.Reserved3)
 	if err != nil {
 		return nil, errors.Wrapf(err, "decoding r.Reserved3 failed")
 	}
 	for index, b := range reserved3ByteArray {
-		report[index+480] = b
+		report[index+504] = b
 	}
 
 	signatureByteArray, err := hex.DecodeString(r.Signature)


johannww:
@hgarvison , would you mind adding a note or a Co-authored-by tag to acknowledge the contribution? It helps with contribution tracking and would mean a lot. Thanks!

KenGordon:
Can you add a "do not use, only intended for testing" comment in front of this please. It will always break when AMD changes fields from padding to useful/real things.

KenGordon (dismissed):
As per my general comment. This cannot be expected to always work in the face of new hardware from AMD, given it was already trying to preserve the padding. Please make a comment to that effect.

I don't think it is useful or sensible to try to create a report rather than use one supplied by AMD hardware. The exception is for fake reports when we might fill out REPORT DATA or HOST DATA but never check the signature just for the purpose of running tests in pools that do not support real SNP hardware.

johannww:
@KenGordon, I do not totally understand the point here. Page 57 of the [AMD spec](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf) states that the the Reserved3 start at 1F8h (504 decimal). So you are saying that this does not make sense for older versions, right?

My intention here is to keep **consistency** between what is deserialized and re-serialized. Let's say I want to to return the report as a json. If I deserialize and re-serialize, I should end up with the same byte array.

Specifically in my case, I need to work with deserialization to verify the passed *ReportData*.

johannww:
Correct me if I am wrong, on the old version, the space these 3 fields occupy are considered reserved. Therefore, they would equally consist of zeroes.

hgarvison:
@johannww I made you an author to the commit and tagged myself as co-author.

KenGordon:
> Correct me if I am wrong, on the old version, the space these 3 fields occupy are considered reserved. Therefore, they would equally consist of zeroes.

The attestation report reserved fields could have been anything previously. Today some say "reserved", some say "reserved must be zero". I personally would not want to be making assumptions about the contents of an object like this except where explicitly specified by AMD.

KenGordon:
> @KenGordon, I do not totally understand the point here. Page 57 of the [AMD spec](https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/specifications/56860.pdf) states that the the Reserved3 start at 1F8h (504 decimal). So you are saying that this does not make sense for older versions, right?
> 
> My intention here is to keep **consistency** between what is deserialized and re-serialized. Let's say I want to to return the report as a json. If I deserialize and re-serialize, I should end up with the same byte array.
> 
> Specifically in my case, I need to work with deserialization to verify the passed _ReportData_.

I think it is ok to de-serialise a report, as you see you need to do that to check report data and all those fields are needed by a relying part to do its work. The use case for serialising one I think is not so clear. We sometimes do it to provide "home made" reports from real templates when running code on non-SNP machines. If you are trying to debug something that deals in these reports on your local machine or you have something that needs to be the same for a CI, or you want a deliberately broken one then this is what you do. However, I would not expect to construct one from the "serialised" version and have the signing still be good. One day it might be, one day it might not. To be confident you would need all fields to be perfectly handler and personally I like to think of the report as an opaque blob with some accessors, not as a thing I can construct. This is due to it being thing signed with some private key we can never have access to.

KenGordon (approved):
Mostly Harmless.

PR #221, opened 2025-06-25
add EC-HSM test for SKR

hgarvison:
add EC-HSM test for SKR

### DIFF
--- a/tests/skr/key.py
+++ b/tests/skr/key.py
@@ -10,7 +10,7 @@ import subprocess
 import tempfile
 
 
-def generate_key():
+def generate_oct_key():
     with tempfile.NamedTemporaryFile() as tmp_key_file:
         print("Generating key file")
         subprocess.check_call(
@@ -55,21 +55,24 @@ def generate_release_policy(attestation_endpoint, host_data):
 
 def deploy_key(
     key_id: str,
+    key_ops: list[str],
     attestation_endpoint: str,
     hsm_endpoint: str,
     key_data: bytes,
     security_policy: str,
+    kty: str = "oct",
 ):
 
     response = requests.put(
         url=f"https://{hsm_endpoint}/keys/{key_id}?api-version=7.4",
         data=json.dumps(
             {
+                # https://learn.microsoft.com/en-us/cli/azure/keyvault/key?view=azure-cli-latest#az-keyvault-key-create
                 "key": {
-                    "kty": "oct-HSM",
+                    "kty": kty, # Key types: EC, EC-HSM, RSA, RSA-HSM, oct, oct-HSM
                     "k": urlsafe_b64encode(binascii.unhexlify(key_data)).decode(),
                     "key_size": 256,
-                    "key_ops": ["encrypt", "decrypt", "wrapKey", "unwrapKey"],
+                    "key_ops": key_ops, # list of permitted JSON web key operations: decrypt, encrypt, export, import, sign, unwrapKey, verify, wrapKey
                 },
                 "hsm": True,
                 "attributes": {
@@ -100,4 +103,4 @@ def deploy_key(
     )
 
     assert response.status_code == 200, response.content
-    print(f"Deployed key {key_id} into the HSM")
+    print(f"Deployed {kty} key {key_id} into the HSM")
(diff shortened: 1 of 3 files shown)


tests/skr/key.py
```diff
@@ -66,10 +68,10 @@ def deploy_key(
         data=json.dumps(
             {
                 "key": {
-                    "kty": "oct-HSM",
+                    "kty": kty,
```
SethHollandsworth: does `kty` need a default value since it's replacing a hardcoded string? It would be nice for a comment to list out the possible values for kty and key__ops as well

tests/skr/test.py
```diff
@@ -205,20 +205,53 @@ def test_skr_http_attest_maa(self):
...
+        if self.attestation_endpoint != "" and self.hsm_endpoint != "":
             # Deploy a Key to the mHSM
             key_id = f"{self.id}-key"
             with open(os.path.join(os.path.realpath(os.path.dirname(__file__)), "policy_skr.rego")) as f:
                 deploy_key(
                     key_id=key_id,
+                    kty="oct-HSM",
+                    key_ops=["encrypt", "decrypt", "wrapKey", "unwrapKey"],
                     attestation_endpoint=self.attestation_endpoint,
                     hsm_endpoint=self.hsm_endpoint,
```
ksayid, line 218: kinda related to seth's comment, what's the difference between kty == "oct" and kty == "oct-HSM" and do we wanna cover both scenarios?

tests/skr/test.py
```diff
@@ -205,20 +205,53 @@ def test_skr_http_attest_maa(self):
...
+                security_policy = generate_release_policy(
+                    attestation_endpoint=self.attestation_endpoint,
+                    host_data=hashlib.sha256(f.read().encode()).hexdigest()
+                )
+                subprocess.run([
+                    "az", "keyvault", "key", "create",
+                    "--id", f"https://{self.hsm_endpoint}/keys/{key_id}",
+                    "--ops", "sign", "verify",
+                    "--kty", "EC-HSM", "--curve", "P-256", "--exportable",
+                    "--policy", f"{security_policy}"],
```
ksayid: nit: no need for f-string on security_policy I think?

tests/skr/test.py
```diff
@@ -232,12 +265,10 @@ def test_skr_http_key_release(self):
...
                 ),
             )
+            key = json.loads(json.loads(skr_response.content.decode())["key"])
             assert skr_response.status_code == 200, skr_response.content.decode()
-            assert json.loads(json.loads(skr_response.content.decode())["key"])["k"] != ""
-            expected_key_ops = {"encrypt", "decrypt", "wrapKey", "unwrapKey"}
-            assert set(json.loads(json.loads(skr_response.content.decode())["key"])["key_ops"]) == expected_key_ops
-        else:
-            print("\nSkipping Key Release test as MAA/mHSM endpoints not provided.\n")
+            assert key["k"] != "" if key["kty"] == "oct" else key["x"] != "" and key["y"] != ""
```
ksayid: could it return kty == "oct-HSM" instead of "oct"?

ksayid: could we also check key_id too?

hgarvison: I checked the response from the key (a part of debug) and it returned oct instead of oct-HSM even though we define the key as oct-HSM, implying that they're the same thing.

tests/skr/test.py
hgarvison: I think they're essentially the same: https://learn.microsoft.com/en-us/java/api/com.azure.security.keyvault.keys.models.keytype?view=azure-java-stable however EC/RSA/oct keys are only available in AKV and the HSM keys are available in akv and mhsm: https://www.cloudpartner.fi/?p=9239

tests/skr/test.py
ksayid: do we want `check=True` for subprocess?
https://csatlas.com/python-subprocess-run-exit-code/#:~:text=done!%27%20)-,check%3DTrue%20Option%20(Shorter%20Code),-But%20the%20shorter

tests/skr/test.py
ksayid: ah gotcha, this seems ok to me. I guess if there's a risk it could spit out oct-HSM, we could check if "oct" is in the kty instead of "oct" == kty if we wanna be careful.

hgarvison: good idea

tests/skr/test.py
hgarvison: I forgot to change this back to check_call instead of run after debug

ksayid (approved):

PR #218, opened 2025-06-19
add attributes file to have git treat .sh and .json files as binary

hgarvison:
add attributes file to have git treat .sh and .json files as binary

https://git-scm.com/docs/gitattributes
https://git-scm.com/docs/gitattributes#_defining_macro_attributes

PR #222, opened 2025-07-07
Updated SKR/ACI example's documentation and streamlined the setup

EricLBaker:
> [!TIP]  
> To view the primary changes from `skr/aci/README.md` without the cluttered `unified diff` view, you can use these links: [old ](https://github.com/microsoft/confidential-sidecar-containers/blob/main/examples/skr/aci/README.md)& [new](https://github.com/microsoft/confidential-sidecar-containers/blob/6ec93deec450f0ad808199e510476ebca07582ad/examples/skr/aci/README.md).

## Summary
- Updated the documentation for the SKR example in the `examples/skr/aci/README.md` file.
- Added a couple small files to help automate credential gathering and populate some `env` variables for the setup.
- For reference - I did multiple runs through this example using the `Premium AKV` SKU.

## Changes
- Broke up blocks of text into actionable items for users to go through
- Unified variable and placeholder names throughout the templates and readme to avoid confusion

## Additions
- CLI command alternatives to most (if not all) actions that previously required using the Azure Portal
- Steps that were missing but required
- Some useful commands and links throughout
- `.env` and `env.sh` files to help streamline the setup. With these, users can now: 
    - Fill out a few variables once in the `.env` file
    - Run `source env.sh` to populate them in their shell
    - Then, they can simply copy paste all CLI commands moving forward with no need to parse/edit each one manually

### DIFF
--- a/examples/skr/aci/.env
+++ b/examples/skr/aci/.env
@@ -1,0 +1,18 @@
+# --------------------------------------------------
+# Fill in variables below, then run `source env.sh`
+# --------------------------------------------------
+
+# Enter names of existing resources or new names to create them:
+MANAGED_ID_NAME="YOUR_MANAGED_ID_NAME"
+RESOURCE_GROUP="YOUR_RESOURCE_GROUP_NAME"
+VAULT_NAME="YOUR_KEYVAULT_NAME"
+KEY_NAME="YOUR_KEY_NAME"
+
+# Choose a valid Azure region (e.g. eastus2, westus2, etc.)
+# To list available regions, run:
+#   az account list-locations --query "[].name" -o tsv
+REGION="YOUR_REGION"
+
+# Don't modify this unless changing the template file's 
+# name or moving it to a different directory.
+TEMPLATE_PATH="aci-arm-template.json"
--- a/examples/skr/aci/README.md
+++ b/examples/skr/aci/README.md
@@ -1,42 +1,87 @@
 # Attestation and Secure Key Release Sidecar ACI Example
 
-## Table of Contents
+<br> 
 
-- [Policy generation](#policy-generation)
-  - [Option 1 - ARM Template](#option-1---arm-template)
-  - [Option 2 - Image-attached fragments](#option-2---image-attached-fragments)
-- [Step by Step Example](#step-by-step-example)
+## 🔖 Table of Contents
 
-In our confidential container group example, we will deploy the skr sidecar along with a set of test containers that exercise and test the REST API.
+- [Attestation and Secure Key Release Sidecar ACI Example](#attestation-and-secure-key-release-sidecar-aci-example)
+  - [🔖 Table of Contents](#-table-of-contents)
+  - [🌐 Overview](#-overview)
+  - [🧰 Policy generation](#-policy-generation)
+  - [🚀 Quickstart -  Running SKR Sidecar on Confidential ACI](#-quickstart----running-skr-sidecar-on-confidential-aci)
+    - [🔗 1. Obtain an Attestation Endpoint](#-1-obtain-an-attestation-endpoint)
+    - [🔐 2. Azure Key Vault (AKV) and User Managed Identity](#-2-azure-key-vault-akv-and-user-managed-identity)
+      - [Useful AKV Commands](#useful-akv-commands)
+      - [2.1 Create a Vault](#21-create-a-vault)
+      - [2.2 Generate a User Managed Identity](#22-generate-a-user-managed-identity)
+      - [2.3 Check User Managed Identity](#23-check-user-managed-identity)
+    - [🪪 3. Populate Image Registry Credentials](#-3-populate-image-registry-credentials)
+    - [🪙 4. Obtain the AAD token](#-4-obtain-the-aad-token)
+    - [📝 5. Fill in Key Information](#-5-fill-in-key-information)
+    - [📄 6. Generate Security Policy](#-6-generate-security-policy)
+    - [📤 7. Import Keys into AKV/mHSM](#-7-import-keys-into-akvmhsm)
+    - [📦 8. Deployment](#-8-deployment)
+    - [📋 9. List and Monitor Deployments](#-9-list-and-monitor-deployments)
 
-- **skr sidecar.** The sidecar’s entry point is /skr.sh which uses the SkrSideCarArgs environment variable to pass the certificate cache endpoint information.
-- **attest/raw test.** The sidecar’s entry point is /tests/skr/attest_client.sh which uses the AttestClientRuntimeData environment variable to pass a blob whose sha-256 digest will be encoded in the raw attestation report as report_data.
-- **attest/maa test.** The sidecar’s entry point is /tests/skr/attest_client.sh which uses two environment variables: (i) AttestClientMAAEndpoint passes the Microsoft Azure Attestation endpoint which will author the attestation token, (ii) AttestClientRuntimeData passes a blob whose sha-256 digest will be encoded in the attestation token as runtime claim.
-- **key/release test.** The sidecar’s entry point is /tests/skr/skr_client.sh which uses the three environment variables: (i) SkrClientKID passes the key identifier of the key to be released from the key vault, (ii) SkrClientAKVEndpoint passes the key vault endpoint from which the key will be released, and (iii) SkrClientMAAEndpoint passes the Microsoft Azure Attestation endpoint shall author the attestation token required for releasing the secret.
-The MAA endpoint shall be the same as the one specified in the SKR policy during the key import to the key vault.
+<br> 
 
-### Policy generation
+## 🌐 Overview 
+In our confidential container group example, we will deploy the SKR sidecar along with a set of test containers that exercise and test the REST API:
+
+| Container             | Entry Point                           | Environment Variables                                                                                                            |
+|---------------------|----------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
+| `skr sidecar`       | `/skr.sh`                              | - `SkrSideCarArgs`: passes the certificate cache endpoint information. |
+| `attest/raw test`   | `/tests/skr/attest_client.sh`          | - `AttestClientRuntimeData`: passes a blob whose `SHA-256` digest will be encoded in the raw attestation report as `report_data`. |
+| `attest/maa test`   | `/tests/skr/attest_client.sh`          | - `AttestClientMAAEndpoint`: passes the Microsoft Azure Attestation endpoint which will author the attestation token.<br>- `AttestClientRuntimeData`: passes a blob whose `SHA-256` digest will be encoded in the attestation token as runtime claim. |
+| `key/release test`  | `/tests/skr/skr_client.sh`             | - `SkrClientKID`: passes the key identifier of the key to be released from the key vault.<br>- `SkrClientAKVEndpoint`: passes the key vault endpoint from which the key will be released.<br>- `SkrClientMAAEndpoint`: passes the Microsoft Azure Attestation (MAA) endpoint shall author the attestation token required for releasing the secret. The MAA endpoint shall be the same as the one specified in the SKR policy during the key import to the key vault. |
+
+For issues during setup, refer to the Troubleshooting guide: [`examples\skr\TROUBLESHOOTING.md`](..\..\skr\TROUBLESHOOTING.md)
+
+<br> 
+
+## 🧰 Policy generation
 
 Deploying a confidential container group requires generating a security policy that restricts what containers can run within the container group.
-To generate security policies, install the Azure `confcom` CLI extension by following the instructions [here under extension examples](https://github.com/Azure/azure-cli-extensions/tree/main/src/confcom/azext_confcom#microsoft-azure-cli-confcom-extension-examples).
 
-There are two options for generating security policies:
+To generate security policies: 
+- [ ] Install the [Azure CLI](https://learn.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest#install)
+- [ ] Then, install the `confcom` CLI extension [here](https://github.com/Azure/azure-cli-extensions/tree/main/src/confcom/azext_confcom#microsoft-azure-cli-confcom-extension-examples) (instructions under 'extension examples').
+
+There are **two options** for generating security policies:
 
 1. The ARM template can be used directly to generate a security policy.
 2. A config file can be used to generate a policy fragment for a particular image (this is called "image-attached fragments" and is useful in cases where stable key release policies are required).
 
-#### Option 1 - ARM Template
+<br>
 
-The following command generates a security policy and automatically injects it into the template.
-Include the `--debug-mode` option so the generated policy allows shelling into container to see the released key in this example. Note that `--debug-mode` is only used as an example and not recommended for production.
+---
+
+<details>
+  <summary><h3>ㅤOption 1:ㅤARM Template </h3></summary>
+
+<br> 
+
+The following command generates a security policy and automatically injects it into the template. <br>
+Include the `--debug-mode` option so the generated policy allows shelling into the container to see the released key in this example.
+
+> ⚠️ **Warning**: `--debug-mode` is only used as an example and **not recommended for production**.
 
 ```shell
-az confcom acipolicygen -a template.json --debug-mode
+az confcom acipolicygen -a aci-arm-template.json --debug-mode
 ```
 
-#### Option 2 - Image-attached fragments
+More information on ARM templates [here](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/overview).
 
-The following command generates a security policy and attaches it to the container in an OCI-compliant registry like Azure Container Registry (ACR). To do this, a copy of the SKR container must be made and pushed to an ACR instance with write access. This can be done with:
+<br>
+
+</details>
+
+---
+
+<details>
+  <summary><h3>ㅤOption 2:ㅤImage-attached fragments </h3></summary>
+
+  The following command generates a security policy and attaches it to the container in an OCI-compliant registry like Azure Container Registry (ACR). To do this, a copy of the SKR container must be made and pushed to an ACR instance with write access. This can be done with:
 
 ```bash
 az acr login -n <my-registry>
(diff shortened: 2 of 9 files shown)


examples/skr/aci/.env
```diff
@@ -0,0 +1,15 @@
...
+# Fill in variables below, then run `source env.sh`
+# --------------------------------------------------
+
+# Enter names of existing resources or new names to create them:
+MANAGED_ID_NAME="YOUR_MANAGED_ID_NAME"
+RESOURCE_GROUP="YOUR_RESOURCE_GROUP_NAME"
+VAULT_NAME="YOUR_KEYVAULT_NAME"
+KEY_NAME="YOUR_KEY_NAME"
+
+# Choose a valid Azure region (e.g., eastus2, westus2, etc.)
```
hgarvison: You could add the command to get the azure regions as a comment here.

examples/skr/aci/.env
```diff
@@ -0,0 +1,15 @@
...
+REGION="YOUR_REGION"
+
+# Don't modify unless changing the template file's name
+TEMPLATE_NAME="aci-arm-template.json"
```
hgarvison: TEMPLATE_PATH in case the user copies these files to their own directory or repo for their own pipeline use?

examples/skr/aci/importkeyconfig.json
```diff
@@ -1,9 +1,9 @@
 {
     "key": {
-        "kid": "<key-vault-key-identifier>",
+        "kid": "<key-vault-key-name-to-create>",
```
hgarvison: I would just leave it at key-vault-key-name to match what we have in the ARM template.

tools/importkey/README.MD
```diff
@@ -1,92 +1,89 @@
-# importkey
```
hgarvison, line 1: I think you need to change the line endings to LF in your VScode

examples/skr/aci/env.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
hgarvison: nit: /bin/sh? https://stackoverflow.com/questions/5725296/difference-between-sh-and-bash

examples/skr/aci/env.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Load variables from .env
+if [ -f ".env" ]; then
```
hgarvison, line 4: Should we add a step for az login before this? Or have a comment/instruction specifying that you need to be logged in?

examples/skr/aci/env.sh
```diff
@@ -0,0 +1,32 @@
...
+export PRINCIPAL_ID=$(az identity show \
+  --name "$MANAGED_ID_NAME" \
+  --resource-group "$RESOURCE_GROUP" \
+  --query principalId -o tsv)
+
+export USER_ID=$(az ad signed-in-user show --query id -o tsv)
+
+# Use this to check everything is populated correctly
+print_env() {
+    echo "Loading environment variables..."
```
hgarvison: nit: I would say something like "Environment variables set..." since this comes after we set them.

examples/skr/aci/README.md
```diff
@@ -1,250 +1,607 @@
-# Attestation and Secure Key Release Sidecar ACI Example
-
-## Table of Contents
```
hgarvison, line 3: Change the line endings to LF so that the diff renders correctly.

examples/skr/aci/env.sh
EricLBaker: We definitely can - this is the behavior when they aren’t logged in though, which I think is pretty intuitive:

image

hgarvison: I'm good with that.

examples/skr/aci/.env
EricLBaker: That makes sense - I left it as the name since I specified [here](https://github.com/microsoft/confidential-sidecar-containers/blob/b46958b151a328c182741bd35dd3509c7136fd50/examples/skr/aci/README.md?plain=1#L518) to run from the root directory of the example (though, I should probably clarify this earlier on in the doc).

That said, I think more adjustments and testing would be required to fully support running this outside the example directory in the way you mentioned - which raises the question: are we aiming to support every possible use case, or is the goal of the `README` simply to demonstrate a working example?

examples/skr/aci/importkeyconfig.json
EricLBaker: Sure thing - I changed it to this to be more explicit about its purpose because a few things were unclear in the docs:
- whether it's an ID or name that's required
- whether the user is supposed to have created a key already, or, if they are supposed to fill in some arbitrary name to create a new one.

Happy to update the placeholder to what you suggested though, or, I can also update the one in the ARM template to match this one.

tools/importkey/README.MD
EricLBaker: Question - Do we want every file to use LF endings? 

If so, I believe we could add something to `.gitattributes` to enforce that in the future

examples/skr/aci/importkeyconfig.json
hgarvison: I would update this one and specify in the readme about the fact that the user comes up with this name.

tools/importkey/README.MD
hgarvison: We decided that previously for consistency in the files so that we don't end up with these diff issues.

I added an attributes file in this PR: https://github.com/microsoft/confidential-sidecar-containers/pull/218 but it seems what I did wasn't sufficient.

examples/skr/aci/.env
hgarvison: The README is to demonstrate a working example, but we've had customers come back with issues because they didn't follow our README. So, I'm just thinking we play it safe in terms of describing things for them in the files, but I would hope they could figure out to run this in the root by running the script.

tools/importkey/README.MD
ksayid: @hgarvison I think for the repo, it should be at `.gitattributes` but if you want to apply it for your local settings, `.git/info/attributes` works 
https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes

tools/importkey/README.MD
```diff
@@ -55,38 +55,36 @@ The authority (MAA endpoint) is also included in the policy so that the vault is
...
+- `-kp` - Path for RSA key in `ASN.1 DER` wrapped in `PKCS8` format.
 
-- -out
+- `-a` - Flag to indicate whether running inside Azure (`default = false`)
 
-    Flag to indicate whether outputing a binary file containing the oct key.
+- `-out` - Flag to indicate whether outputting a binary file containing the oct key.
     For RSA keys, the oct key is derived from the RSA's secret exponent.
-    The key derivation information may be passed via the key_derivation attribute (salt passed as a hexstring and label passed as a string).
+    The key derivation information may be passed via the `key_derivation` attribute (salt passed as a hexstring and label passed as a string).
```
hgarvison: Where does this attribute get set?

tools/importkey/README.MD
hgarvison: @ksayid The wording from the Git website implies that .gitattributes would affect other repos which is why I went with the info/attributes (https://git-scm.com/docs/gitattributes).

Additionally, their hierarchy of checks implies that without .gitattributes, info/attributes should be checked https://git-scm.com/docs/gitattributes#_defining_macro_attributes.

We can move it though.

examples/skr/aci/env.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/sh
```
ksayid, line 1: What do you think about adding `set -euo pipefail` to make sure the script exits immediately on errors? I think right now, it might continue. For example, if an Azure CLI command fails, variables could remain unset, which might lead to other issues later in the script

examples/skr/aci/README.md
```diff
@@ -1,42 +1,87 @@
 # Attestation and Secure Key Release Sidecar ACI Example
 
-## Table of Contents
+<br>
```
hgarvison, line 3: Per Copilot, I don't think you need these breaks

examples/skr/aci/README.md
```diff
@@ -1,42 +1,87 @@
 # Attestation and Secure Key Release Sidecar ACI Example
 
-## Table of Contents
+<br> 
+
+## 🔖 Table of Contents
+
+- [Attestation and Secure Key Release Sidecar ACI Example](#attestation-and-secure-key-release-sidecar-aci-example)
```
hgarvison, line 7: Do we need to add these first two links? Seems redundant since we're in that section already.

examples/skr/aci/README.md
```diff
@@ -1,42 +1,87 @@
...
+| `attest/raw test`   | `/tests/skr/attest_client.sh`          | - `AttestClientRuntimeData`: passes a blob whose `SHA-256` digest will be encoded in the raw attestation report as `report_data`. |
+| `attest/maa test`   | `/tests/skr/attest_client.sh`          | - `AttestClientMAAEndpoint`: passes the Microsoft Azure Attestation endpoint which will author the attestation token.<br>- `AttestClientRuntimeData`: passes a blob whose `SHA-256` digest will be encoded in the attestation token as runtime claim. |
+| `key/release test`  | `/tests/skr/skr_client.sh`             | - `SkrClientKID`: passes the key identifier of the key to be released from the key vault.<br>- `SkrClientAKVEndpoint`: passes the key vault endpoint from which the key will be released.<br>- `SkrClientMAAEndpoint`: passes the Microsoft Azure Attestation (MAA) endpoint shall author the attestation token required for releasing the secret. The MAA endpoint shall be the same as the one specified in the SKR policy during the key import to the key vault. |
+
+For issues during setup, refer to the Troubleshooting guide: [`examples\encfs\TROUBLESHOOTING.md`](..\..\..\examples\encfs\TROUBLESHOOTING.md)
```
hgarvison: This points to the ENCFS troubleshooting guide instead of the SKR troubleshooting guide

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+<br>
 
-Here is an example of running skr sidecar on confidential ACI.
-The MAA endpoint is the value of env var [`SkrClientMAAEndpoint`](aci-arm-template.json?plain=1#L55).
-The managed HSM instance endpoint corresponds to [`SkrClientAKVEndpoint`](aci-arm-template.json?plain=1#L59).
-We will also import a key into managed HSM under the name [`doc-sample-key-release`](aci-arm-template.json?plain=1#L64)
+## 🚀 Quickstart -  Running SKR Sidecar on Confidential ACI
 
-#### 1. Obtain an Attestation Endpoint
+The following is an example of running SKR sidecar on confidential ACI.
```
hgarvison: Capitalize confidential for consistency with title?

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+- The MAA endpoint is the value of env var [`SkrClientMAAEndpoint`](aci-arm-template.json?plain=1#L55).
+- The managed HSM instance endpoint corresponds to [`SkrClientAKVEndpoint`](aci-arm-template.json?plain=1#L59).
+- We will also import a key into managed HSM under the name [`doc-sample-key-release`](aci-arm-template.json?plain=1#L64).
```
hgarvison: we don't name the key that in the arm template

ksayid (approved):

tools/importkey/README.MD
EricLBaker: @ksayid - Moved `info/attributes` to the root dir and renamed to `.gitattributes`

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+> If you are granted new permissions, use the following to refresh credentials: `az account clear && az login`
 
-At this point, the `aci-arm-template.json` file should be filled out except for the `ccepolicy` field.
-After installing the [Azure `confcom` CLI extension](#policy-generation), run the following command to generate the security policy and include the `--debug-mode` option so that the policy allows users to shell into the container.
+
+<br>
+
+After setting up an Azure Key Vault resource:
+
+- [ ]  You can create a user-assigned managed identity [here](https://learn.microsoft.com/en-us/entra/identity/managed-identities-azure-resources/how-manage-user-assigned-managed-identities?pivots=identity-mi-methods-azp) that will be attached to the container group so that the containers have the correct access permissions to Azure services and resources.
```
hgarvison: Instead of 'here,' either link the 'create a user-managed identity' part or say 'by following these instructions'

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+  ```shell
+  az identity create -n "$MANAGED_ID_NAME" -g "$RESOURCE_GROUP" -l "$REGION"
+  ```
+  More information about creating identities can be found [here.](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/)
+
+<br>
+
+If using ***standard*** or ***premium*** AKV:
+
+- [ ] Assign the `Key Vault Crypto Service Release User` role to your managed identity <br>
```
hgarvison, line 379: I don't think this needs the br

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+
+If you already have a user-assigned managed identity with the appropriate access permissions:
+
+- [ ] Run the following command to list the managed identities for a `RESOURCE_GROUP`:
+
+  ```shell
+  az identity list -g "$RESOURCE_GROUP"
+  ```
+
+  - [ ] Or, if you know the `MANAGED_ID_NAME` **and** the `RESOURCE_GROUP`, run:
```
hgarvison: Put this at the same level as the 'Run' step

examples/skr/aci/README.md
```diff
@@ -122,105 +167,362 @@ Example output:
...
+  - [ ] Run the following command to get the full managed identity, and replace [`full-path-to-managed-identity-with-right-permissions-to-key-vault`](aci-arm-template.json#L22) with the output:
+
+    ```shell
+    az identity show -n "$MANAGED_ID_NAME" -g "$RESOURCE_GROUP" --query id -o tsv
+    ```
+  - [ ] Remove or fill out `LogFile` and `LogLevel` (under `environmentVariables`)
+
+<br>
+
+### 📄 6. Generate Security Policy
```
hgarvison, line 496: May want to have a link back to the overview at the top jic

examples/skr/aci/README.md
```diff
@@ -229,22 +531,77 @@ https://<mhsm-name>.managedhsm.azure.net/keys/doc-sample-key-release/8659****0cd
...
+    level=debug msg=Releasing key blob: {doc-sample-key-release}
+    ```
+
+Alternatively, you can:
+
+- [ ] Shell into the container `test-skr-client-hsm-skr` and the released key is in `keyrelease.out`.
+
+<br>
+
+### 📋 9. List and Monitor Deployments
```
hgarvison, line 584: I would provide a link to the readme for these commands

examples/skr/aci/env.sh
EricLBaker: I like it and I tested something similar earlier using `set -e`, but I wish we could find an alternative that doesn't cause the terminal to crash and exit on failure. It's not conducive to debugging if you don't see the error before the terminal closes. It seems this approach only works nicely for `bash` - when I run it in `zsh` / `wsl`, that's where the issues pop up. 

I also don't think it's necessary to have to catch these for this specific script, as it's more of a "nice to have". When prompted to run `print_env`, users will see if the variables were populated or not.

examples/skr/aci/README.md
EricLBaker: Agreed - The table of contents is auto-generated by a markdown plugin I use, I can try to remove those manually though.

_Update_: They came back after removing lol. Hard to justify removing the plugin over that though, it keeps all the section labels updated in a two-way fashion which is really convenient.

examples/skr/aci/README.md
EricLBaker: Nice catch!

tools/importkey/README.MD
ksayid: @hgarvison, ah yeah that phrasing from the site is confusing. I just looked into it
* "distributed to other repositories" means .gitattributes travels with this repository when cloned or forked
* every git repo we clone or create has a hidden `.git` folder in the root of the repository. If you place your attributes file at `.git/info/attributes` (inside this hidden folder) and don't commit or push it, the rules defined there will only affect your local clone.  so, if the line-ending or binary-file rules are in this file locally, your own git diff commands will respect those settings without impacting others / appearing in GitHub's diff viewer online. but leaving the file at `info/attributes` (so at the repository root, outside the hidden .git folder) won't do anything, because Git never checks there by default.

tools/importkey/README.MD
EricLBaker: Not sure, I didn't change or add any content to this file, just updated the formatting and typos - It says you are the one who wrote this before my current edit: 
 
image

hgarvison: Ah that PR was just me make a newline after every sentence lol and this file was made before we took over the repo. I think it's in the importconfig file and we should specify it. Plus, this sentence shouldn't be attached to the -out flag.

examples/skr/aci/README.md
EricLBaker: Copilot is great but not very reliable when it comes to advice on formatting.

Without using breaks, all the content runs together and is not very digestible.

image

examples/skr/aci/README.md
EricLBaker: Not sure I see what you mean - Why would we need a link from section 6 to the overview?

examples/skr/aci/README.md
EricLBaker: You mean to the documentation, like this?

> For the most accurate and up-to-date info on these commands, refer to:
>
> - [`az container`](https://learn.microsoft.com/en-us/cli/azure/container?view=azure-cli-latest&utm_source=chatgpt.com)
> - [`az deployment group`](https://learn.microsoft.com/en-us/cli/azure/deployment/group?view=azure-cli-latest&utm_source=chatgpt.com)

ksayid (approved):

examples/skr/aci/README.md
hgarvison: I meant the section under it on policy generation.

PR #224, opened 2025-07-09
add combined attestation endpoint to grpc endpoints

hgarvison:
added grpc endpoint `GetAttestationData` with the functionality of the combined attestation http endpoint to grpc endpoints


stevendongatmsft (approved):

SethHollandsworth (approved):
lgtm

tests/skr/test.py
```diff
@@ -295,6 +295,37 @@ def test_skr_grpc_get_report(self):
...
 
+    def test_skr_grpc_get_attestation_data(self):
+
+        input_report_data = b"EXAMPLE"
+        response = requests.get(
+            f"http://{self.skr_ip}:8000/get_attestation_data",
+            headers={
+                "Content-Type": "application/json",
+            },
+            data=json.dumps(
```
ksayid, line 306: might be a silly question but if we're sending a body, would a POST request make more sense?

cmd/skr/README.md
```diff
@@ -144,6 +144,13 @@ Users can optionally provide `reportDataHexString` and the input will show under
 grpcurl -v -plaintext -d '{"reportDataHexString":""}' 127.0.0.1:50000  keyprovider.KeyProviderService.GetReport
 ```
 
+The `GetAttestationData` method of the KeyProviderService is used to get the combined attestation data similar to the HTTP `attest/combined` endpoint.
+Users can optionally provide `b64RuntimeDataString` to prevent replay attacks by guaranteeing the freshness of the attestation report.
+
+```bash
+grpcurl -v -plaintext -d '{"b64RuntimeDataString":""}' 127.0.0.1:50000  keyprovider.KeyProviderService.GetAttestationData
```
ksayid: unrelated nit: I think we say the default port for HTTP is 8080 in para 2, maybe we can also say something about the default port for gRPC?

tests/skr/test.py
hgarvison: I agree, but I was just copying what was existing for the other tests lol

pkg/grpc/keyprovider/keyprovider.proto
```diff
@@ -20,10 +20,22 @@ message keyProviderGetReportOutput {
   string reportHexString = 1;
 }
 
+message keyProviderGetAttestationDataInput {
```
ksayid: just learned about the protobuf style guide from Google: https://protobuf.dev/programming-guides/style/ 

I think the existing code didn't really follow it. this PR is consistent with the existing code, so I think it's all good, but just something to be aware of

cmd/skr/README.md
hgarvison: It doesn't specify HTTP in that because the default port for both is 8080

pkg/grpc/keyprovider/keyprovider.proto
hgarvison: I didn't write the existing code, so I was just staying consistent with it, but it's an easy fix.

cmd/skr/README.md
ksayid: then should the examples be using grpc examples use 8080?

pkg/grpc/keyprovider/keyprovider.proto
ksayid: Definitely, just wanted to share it for reference

ksayid (approved):
LGTM

cmd/skr/README.md
hgarvison: I can add a note that says we they are for an example with a port specified to be 50000

ksayid (approved):

PR #227, opened 2025-08-13
Do not log access tokens and keys in debug mode

micromaomao:
We should not log sensitive materials regardless of the log level. Such logging is surprising and can end up leaking those information if the log is made available in unprotected places or turned on accidentally.  This PR redacts the tokens and removes some log calls with keys.

Example logs from skr:
```
level=info msg=Generating MAA Report Data...
level=info msg=Decoding UVM encoded security policy...
level=info msg=Running inside SNP VM, using real attestation report fetcher...
level=info msg=Generating MAA Report Data...
level=debug msg=   reportData: 5c467494b10092efeefe53f39b795c110e7d4e300a114baff0919c148af5eac90000000000000000000000000000000000000000000000000000000000000000
level=info msg=Fetching Attestation Report...
level=info msg=Deserializing Attestation Report...
level=debug msg=SNP Report Reported TCB: 15787368493747273732
Cert Chain TCBM Value: 15787368493747273732

level=info msg=Comparing TCB values...
level=info msg=TCB values match, using cached cert chain...
level=info msg=Retrieving MAA token...
level=info msg=Constructing MAA Attestation Request...
level=info msg=Marshalling MAA Endorsement...
level=debug msg=MAA Request: {"report":"eyJTbnBSZXBvcnQiOiJBd0FBQUFJQUFBQWZBQU1BQUFBQUFBRUFBQUFBQUFBQUFBQUFBQUFBQUFBQ0FBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBRUFBQUFFQUFBQUFBQVkyeVVBQUFBQUFBQUFBQUFBQUFBQUFBQmNSblNVc1FDUzctNy1VX09iZVZ3UkRuMU9NQW9SUzZfd2tad1VpdlhxeVFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFBQUFYLTdqRFcxLUdpb <snip...>

level=debug msg=Posting MAA Attestation Request to https://confidentialsidecars.weu.attest.azure.net/attest/SevSnpVM?api-version=2020-10-01
level=info msg=Unmarshalling MAA Attestation Response...
level=debug msg=MAA Token (signature redacted): eyJhbGciOiJSUzI1NiIsImprdSI6Imh0dHBzOi8vY29uZmlkZW50aWFsc2lkZWNhcnMud2V1LmF0dGVzdC5henVyZS5uZXQvY2VydHMiLCJraWQiOiJkUktoK2hCY1dVZlFpbVNsM0l2NlpoU3RXM1RTT3QwVGh3aVRnVVVxWkFvPSIsInR5cCI6IkpXVCJ9.eyJleHAiOjE3NTUyMTIwNzgsImlhdCI6MTc1NTE4MzI3OCwiaXNzIjoiaHR...<snip>...S4wIn0.***
```

Example logs from remotefs:
```
level=info msg=2 filesystems to mount:
level=info msg=  /mnt/remote/encfs-16970194319-1795--debug-blob1 -> https://cacisidecarsstorage.blob.core.windows.net/container/encfs-16970194319-1795--debug-blob1 (read-write: true)
level=debug msg=    keyBlob: keyBlob{ KID: encfs-16970194319-1795--debug-key, KTY: , KeyOps: [], Authority: {Endpoint:confidentialsidecars.weu.attest.azure.net TEEType: APIVersion:}, AKV: { Endpoint: cacisidecars.managedhsm.azure.net, APIVersion:  } }
level=debug msg=    KeyDerivationBlob: {Salt: Label:}
level=info msg=  /mnt/remote/encfs-16970194319-1795--debug-blob2 -> https://cacisidecarsstorage.blob.core.windows.net/container/encfs-16970194319-1795--debug-blob2 (read-write: false)
level=debug msg=    keyBlob: keyBlob{ KID: encfs-16970194319-1795--debug-key, KTY: , KeyOps: [], Authority: {Endpoint:confidentialsidecars.weu.attest.azure.net TEEType: APIVersion:}, AKV: { Endpoint: cacisidecars.managedhsm.azure.net, APIVersion:  } }
level=debug msg=    KeyDerivationBlob: {Salt: Label:}
```

### DIFF
--- a/cmd/remotefs/azurefs.go
+++ b/cmd/remotefs/azurefs.go
@@ -266,7 +266,7 @@ func releaseRemoteFilesystemKey(tempDir string, keyDerivationBlob common.KeyDeri
 			return "", errors.Wrapf(err, "failed to derive oct key")
 		}
 
-		logrus.Debugf("Symmetric key %s (salt: %s label: %s)", hex.EncodeToString(octetKeyBytes), keyDerivationBlob.Salt, labelString)
+		logrus.Debugf("Symmetric key salt: %s label: %s", keyDerivationBlob.Salt, labelString)
 	default:
 		return "", errors.Wrapf(err, "key type %s not supported", jwKey.KeyType())
 	}
--- a/cmd/remotefs/main.go
+++ b/cmd/remotefs/main.go
@@ -112,8 +112,27 @@ func main() {
 		logrus.Fatalf("Failed to unmarshal base64 string: %s\n%s", err.Error(), ERROR_STRING)
 	}
 
+	logrus.Infof("%d filesystems to mount:", len(info.AzureFilesystems))
+
 	// populate missing attributes in KeyBlob
-	for i := range info.AzureFilesystems {
+	for i, fs := range info.AzureFilesystems {
+		logrus.Infof("  %s -> %s (read-write: %t)", fs.MountPoint, fs.AzureUrl, fs.ReadWrite)
+		// KeyBlob only contains the identifier of the key, so we can print it,
+		// but we need to ensure we do not print any AKV tokens, if it has any.
+		// SafeString() handles that.
+		logrus.Debugf("    keyBlob: %s", fs.KeyBlob.SafeString())
+		// KeyDerivationBlob contains the salt and label used to derive the key,
+		// but the actual HKDF derivation requires the RSA private key, acquired
+		// via key release, and so we can safely print it since it won't expose
+		// the derived key.
+		logrus.Debugf("    KeyDerivationBlob: %+v", fs.KeyDerivationBlob)
+		if fs.RawKeyHexString != "" {
+			if allowTestingWithRawKey {
+				logrus.Debug("    RawKeyHexString provided, using it for testing and skipping skr")
+			} else {
+				logrus.Warnf("filesystem %s: provided RawKeyHexString will be ignored as this image does not allow testing with raw key input.", fs.MountPoint)
+			}
+		}
 		// set the api versions and the tee type for which the authority will authorize secure key release
 		info.AzureFilesystems[i].KeyBlob.AKV.APIVersion = "api-version=7.4"
 		info.AzureFilesystems[i].KeyBlob.Authority.APIVersion = "api-version=2020-10-01"
--- a/cmd/remotefs/main.go
+++ b/cmd/remotefs/main.go
@@ -120,8 +120,6 @@ func main() {
 		info.AzureFilesystems[i].KeyBlob.Authority.TEEType = "SevSnpVM"
 	}
 
-	logrus.Debugf("JSON = %+v", info)
-
 	err = MountAzureFilesystems(tempDir, info)
 	if err != nil {
 		logrus.Fatalf("Failed to mount filesystems: %s\n%s", err.Error(), ERROR_STRING)
--- a/pkg/common/token.go
+++ b/pkg/common/token.go
@@ -4,7 +4,10 @@
 package common
 
 import (
+	"encoding/base64"
 	"encoding/json"
+	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 )
@@ -59,4 +62,38 @@ func GetToken(resourceId string, i Identity) (r TokenResponse, err error) {
 	}
 
 	return r, nil
+}
+
+// Remove the signature from a MAA token, but leaving the information JSON
+// intact.
+//
+// When loglevel is "debug" or higher, we log the received MAA token. To avoid
+// leaking them, safely redact the token by removing the signature.
+//
+// This function also checks that it is indeed a MAA token, and will redact the
+// whole string if it is not.
+func RedactMAAToken(token string) string {
+	// JWT consists of three parts: header, payload, and signature, separated by
+	// dots.  We check the issuer in the payload and remove the last part.
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return "<redacted invalid token: not a JWT>"
+	}
+	decodedPayload, err := base64.RawURLEncoding.DecodeString(parts[1])
+	if err != nil {
+		return "<redacted invalid token: failed to decode payload>"
+	}
+	var payload map[string]interface{}
+	err = json.Unmarshal(decodedPayload, &payload)
+	if err != nil {
+		return "<redacted invalid token: failed to unmarshal payload>"
+	}
+	issuer, ok := payload["iss"].(string)
+	if !ok {
+		return "<redacted invalid token: invalid issuer>"
+	}
+	if !strings.HasSuffix(issuer, ".attest.azure.net") {
+		return fmt.Sprintf("<redacted token with issuer %s>", issuer)
+	}
+	return strings.Join([]string{parts[0], parts[1], "***"}, ".")
 }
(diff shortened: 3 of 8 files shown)


pkg/common/token.go
```diff
@@ -60,3 +61,17 @@ func GetToken(resourceId string, i Identity) (r TokenResponse, err error) {
...
+// Remove the signature from a JWT.
+//
+// This is used to safely debug log the information within a MAA token. If an
+// invalid token is passed in that does not have 3 parts, it is returned as-is.
+func RedactToken(token string) string {
+	// JWT consists of three parts: header, payload, and signature, separated by
+	// dots.  We remove the last part.
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return token
```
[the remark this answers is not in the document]
micromaomao: Won't fix, so that in case we get anything weird from MAA, we see it.

pkg/common/token.go
```diff
@@ -60,3 +61,14 @@ func GetToken(resourceId string, i Identity) (r TokenResponse, err error) {
...
 	return r, nil
 }
+
+// Remove the signature from a JWT
+func RedactToken(token string) string {
+	// JWT consists of three parts: header, payload, and signature, separated by
+	// dots.  We remove the last part.
+	parts := strings.Split(token, ".")
+	if len(parts) != 3 {
+		return token
```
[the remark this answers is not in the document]
KenGordon: I agree with Copilot.

pkg/common/token.go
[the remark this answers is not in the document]
KenGordon: In that case we will need to build a special test. If our code messes up and the thing passed in here is actually a token but fails your check then you will still log it. I think that we will be criticised for this code path.

pkg/common/token.go
```diff
@@ -60,3 +61,17 @@ func GetToken(resourceId string, i Identity) (r TokenResponse, err error) {
 
 	return r, nil
 }
+
+// Remove the signature from a JWT.
+//
+// This is used to safely debug log the information within a MAA token. If an
```
KenGordon: "safely debug log" - Locally this is regardless of logging level. And if you want to talk about that then maybe "When loglevel is "debug" or higher, to avoid leaking tokens, safely redact them by removing the signature." 

To be sure we don't get pushback on this and it is worth while, can you decode such a token, see that the claims are available and maybe add it as a comment?

cmd/remotefs/azurefs.go
```diff
@@ -266,7 +266,7 @@ func releaseRemoteFilesystemKey(tempDir string, keyDerivationBlob common.KeyDeri
 			return "", errors.Wrapf(err, "failed to derive oct key")
 		}
 
-		logrus.Debugf("Symmetric key %s (salt: %s label: %s)", hex.EncodeToString(octetKeyBytes), keyDerivationBlob.Salt, labelString)
+		logrus.Debugf("Symmetric key: (salt: %s label: %s)", keyDerivationBlob.Salt, labelString)
```
KenGordon: There is no longer a key after "key:" which might confuse someone reading this log in a "was the key empty?" sort of a way.

cmd/remotefs/main.go
```diff
@@ -120,8 +120,6 @@ func main() {
 		info.AzureFilesystems[i].KeyBlob.Authority.TEEType = "SevSnpVM"
 	}
 
-	logrus.Debugf("JSON = %+v", info)
```
KenGordon, line 123: I wonder if we should log something high level per filesystem, even just the number of filesystems.

KenGordon (changes requested):
MInor asks.

pkg/common/token.go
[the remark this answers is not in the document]
micromaomao: Actually, on second thought, I will take the copilot's suggestion and just return `"<redacted invalid token>"` instead, since this token is parsed from the response JSON, and if the response is invalid, we're not getting here anyway.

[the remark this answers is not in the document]
micromaomao: Also added a quick test

pkg/skr/skr.go
```diff
@@ -117,7 +116,7 @@ func SecureKeyRelease(identity common.Identity, certState attest.CertState, skrK
 		return nil, errors.Wrapf(err, "releasing the key %s failed", skrKeyBlob.KID)
 	}
 
-	logrus.Debugf("Key Type: %s Key %v", kty, keyBytes)
+	logrus.Debugf("Key Type: %s", kty)
```
hgarvison: Can we add something to say that the key was successfully released? I'm pretty sure the key type is logged in several places before this so this log does nothing to tell us where in the code we are if we need to debug something.

pkg/common/token.go
```diff
@@ -60,3 +63,37 @@ func GetToken(resourceId string, i Identity) (r TokenResponse, err error) {
...
+	var payload map[string]interface{}
+	err = json.Unmarshal(decodedPayload, &payload)
+	if err != nil {
+		return "<redacted invalid token: failed to unmarshal payload>"
+	}
+	issuer, ok := payload["iss"].(string)
+	if !ok {
+		return "<redacted invalid token: invalid issuer>"
+	}
+	if !strings.HasSuffix(issuer, ".attest.azure.net") {
```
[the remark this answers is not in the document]
micromaomao: "malicious tokens" is not a concern.

pkg/skr/skr.go
micromaomao: Changed to:
```
	logrus.Debugf("Successfully released key with type: %s", kty)
```

cmd/remotefs/main.go
micromaomao: Added in latest commit and added a example to the PR description

cmd/remotefs/azurefs.go
micromaomao: Changed to:
```
		logrus.Debugf("Symmetric key salt: %s label: %s", keyDerivationBlob.Salt, labelString)
```

pkg/common/token.go
micromaomao: This function is now updated and renamed to RedactMAAToken. It will only return the payload in the token if it is a MAA token.

hgarvison (approved):

cmd/remotefs/main.go
```diff
@@ -112,16 +112,26 @@ func main() {
 		logrus.Fatalf("Failed to unmarshal base64 string: %s\n%s", err.Error(), ERROR_STRING)
 	}
 
+	logrus.Infof("%d filesystems to mount:", len(info.AzureFilesystems))
+
 	// populate missing attributes in KeyBlob
-	for i := range info.AzureFilesystems {
+	for i, fs := range info.AzureFilesystems {
+		logrus.Infof("  %s -> %s (read-write: %t)", fs.MountPoint, fs.AzureUrl, fs.ReadWrite)
+		logrus.Debugf("    keyBlob: %s", fs.KeyBlob.SafeString())
```
KenGordon, line 123: Can you say why this is safe explicitly in a comment. Maybe that is what SafeString does. Why is it safe to print KeyDerivationBlob? Can I make the same key again if I know that? Tell us, you probably thought about why it was ok. Save the rest of us having to figure it out.

KenGordon (approved):

cmd/remotefs/main.go
micromaomao: Updated - can you review and approve again?

MahatiC (approved):

PR #228, opened 2025-08-14
Clean up encfs test container group

micromaomao:
Not sure if this is intentional, given that the skr test does clean it up?


DomAyre:
Yeah I don't remember, but could easily be from development and I forgot to clean up

DomAyre (approved):

PR #231, opened 2025-08-26
Update to use latest version (v2.12) with cleaner logging.

KenGordon:
Signed-off-by: Ken Gordon <Ken.Gordon@microsoft.com>

### DIFF
--- a/examples/encfs/aci-arm-template.json
+++ b/examples/encfs/aci-arm-template.json
@@ -44,7 +44,7 @@
                   "value": "<optional-loglevel-trace-debug-info-warning-error-fatal-panic>"
                 }
               ],
-              "image": "mcr.microsoft.com/aci/encfs:2.11",
+              "image": "mcr.microsoft.com/aci/encfs:2.12",
               "securityContext": {
                 "privileged": true
               },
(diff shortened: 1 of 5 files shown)


examples/encfs/aci-arm-template.json
```diff
@@ -44,7 +44,7 @@
                   "value": "<optional-loglevel-trace-debug-info-warning-error-fatal-panic>"
                 }
               ],
-              "image": "mcr.microsoft.com/aci/encfs:2.11",
+              "image": "mcr.microsoft.com/aci/encfs:2.12",
```
hgarvison, line 47: Can you do the same for the ENCFS examples? We made 2.12 for that too. You may also need to update the tests/pipeline versions

KenGordon: I thought I did. Where is it 2.11 still?

hgarvison: Oops I had a brain fart and misread all these as skr. You're good. Though I would update your pipelines in the other repos to point to the newest versions

hgarvison (approved):

examples/encfs/aci-arm-template.json
KenGordon: No problem.

PR #232, opened 2025-09-11
Send a user-agent for MAA requests

micromaomao:


cmd/skr/main.go
```diff
@@ -136,6 +137,22 @@ func main() {
...
+		// the default user agent (i.e. the subscription ID) in a separate
+		// goroutine to not delay server startup.
+		identity := info.Identity // Make a copy
+		go func() {
+			ua := getDefaultMAAUserAgent(identity)
+			common.MAAClientUserAgent = ua
+			logrus.Infof("Successfully fetched token, setting default MAA User-Agent to: %s", ua)
+		}()
+	} else {
+		logrus.Infof("Using provided MAA User-Agent: %s", common.MAAClientUserAgent)
```
KenGordon: Would this be clearer as "Using provided string %s as User-Agent for request to MAA"?

cmd/skr/main.go
```diff
@@ -216,3 +233,46 @@ func setupServer(certState *attest.CertState, identity *common.Identity, uvmInfo
 		logrus.Fatalf("Failed to start HTTP server: %v\n%s", err, skr.ERROR_STRING)
 	}
 }
+
```
KenGordon, line 245: Could we have a comment saying what this function is intended to do and maybe (unless elsewhere) why we want that?

KenGordon: And if there is any documentation about magic strings like "xms_az_rid" a link would be nice. If not a comment saying "obtained by inspection" will be a valuable hint to people in the future when it stops a day after a new style of MI is rolled out :)

pkg/common/http.go
```diff
@@ -61,6 +62,9 @@ func HTTPPRequest(httpType string, uri string, jsonData []byte, authorizationTok
 	if authorizationToken != "" {
 		req.Header.Add("Authorization", "Bearer "+authorizationToken)
 	}
+	for k, v := range extraHeaders {
+		req.Header.Add(k, v)
+	}
```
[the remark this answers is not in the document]
KenGordon: I agree. And even if there is a convoluted reason why that can never happen, I still want to be certain with a test.

cmd/skr/main.go
```diff
@@ -216,3 +233,46 @@ func setupServer(certState *attest.CertState, identity *common.Identity, uvmInfo
...
+func getDefaultMAAUserAgent(identity common.Identity) string {
+	token, err := skr.GetKeyvaultAccessToken(false, identity)
```
KenGordon: Explain this is pragmatic.

cmd/skr/main.go
micromaomao: Updated

KenGordon (approved):

pkg/common/http.go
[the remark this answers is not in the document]
micromaomao: I missed this but in golang `for ... := range nil` does nothing
image

hgarvison:
Can you add tests for this (unit and/or update the pipeline tests)? Also, can we add documentation for this (and update the examples if possible)?

micromaomao:
@copilot Can you add tests for this (unit and/or update the pipeline tests)? Also, can we add documentation for this (and update the examples if possible)?

micromaomao:
I kept thinking I will add some tests/docs at some point, however this has never made it high enough in my priority list :(
(it's also probably a bit tricky to test since one would need to mock the request function to check whether the header is actually sent to MAA)

@KenGordon @hgarvison Can we merge this anyway so that we ensure this gets included in the next skr release? If I find some time I can make a new PR for the tests.

KenGordon (approved):
As discussed.

PR #237, opened 2025-11-06
Fix README for attest package

dwhjames:
`pkg/attest/attestation_report_fetcher.go` implements in pure Go what `tools/get-snp-report` implements in C.


dwhjames:
Related to this, there are some other places that could be simplified, e.g.,

https://github.com/microsoft/confidential-sidecar-containers/blob/c2470b59568ec6f2d8e0bb136be8cb2dafc744e4/docker/skr/Dockerfile.skr#L5

(`get-snp-report` etc. are built and included in the published artifact, but unused.

dwhjames:
@microsoft-github-policy-service agree company="Gemini"

PR #240, opened 2025-12-12
Fix empty subscription ID in logs

micromaomao:


KenGordon (approved):
Isn't this a secret/value that comes from github - kind of like the previous conversations about these?

micromaomao:
> Isn't this a secret/value that comes from github - kind of like the previous conversations about these?

We can also take it out of this .env file, but the advantage of having it be in here is that someone running this locally can `.`-source the env file and run tests via c-aci-testing.

PR #244, opened 2026-02-02
Bump go dependencies

micromaomao:


KenGordon (approved):
Do you need to update version numbers of the sidecar all over the place?

micromaomao:
> Do you need to update version numbers of the sidecar all over the place?

Forgot to check this, but it turns out the only SKR version number this repo references is in the examples. I've created another PR to bump these, to be merged after the image is released (you can approve now):

https://github.com/microsoft/confidential-sidecar-containers/pull/245

There are no other version numbers to bump outside examples/, so I will merge this now

PR #248, opened 2026-06-15
Add secure cert release capability

canfikret:
This is to support secure cert release from AKV, which is a private preview feature.

### DIFF
--- a/internal/httpginendpoints/httpginendpoints.go
+++ b/internal/httpginendpoints/httpginendpoints.go
@@ -35,7 +35,7 @@ type RawAttestData struct {
 	RuntimeData string `json:"runtime_data" binding:"required"`
 }
 
-type KeyReleaseData struct {
+type releaseData struct {
 	// MAA endpoint which acts as authority to the key that needs to be released
 	MAAEndpoint string `json:"maa_endpoint" binding:"required"`
 	// AKV endpoint from which the key is released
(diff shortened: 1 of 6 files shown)


internal/httpginendpoints/httpginendpoints.go
```diff
@@ -340,6 +340,62 @@ func PostKeyRelease(c *gin.Context) {
...
+		APIVersion:  "api-version=7.6-preview.1",
+		BearerToken: newCertReleaseData.AccessToken,
+	}
+
+	logrus.Debugf("Setting AKV API version to 7.6-preview.1")
+
+	maa := common.MAA{
+		Endpoint:   newCertReleaseData.MAAEndpoint,
+		TEEType:    "SevSnpVM",
+		APIVersion: "api-version=2020-10-01",
```
KenGordon, line 363: Does this need to be updated as this is a new feature?

pkg/common/akv.go
```diff
@@ -325,37 +350,44 @@ func (akv AKV) ImportPlaintextKey(key interface{}, releasePolicy ReleasePolicy,
 // the release policy. ReleaseKey uses the private key to locally unwrap the released secrets.
 // The private key is kept within the utility VM and hence is isolated with hardware-based
 // guarantees.
-func (akv AKV) ReleaseKey(maaTokenBase64 string, kid string, privateWrappingKey *rsa.PrivateKey) (_ []byte, _ string, _ []string, err error) {
+func (akv AKV) ReleaseKey(maaTokenBase64 string, kid string, privateWrappingKey *rsa.PrivateKey, isCertRelease bool) (_ []byte, _ string, err error) {
```
KenGordon, line 353: Why is this changing the returned values? Is it just "cleaning up while here" or was it ever really returning anything?

KenGordon (commented):
Let's have Tingmao look at it properly. She is in the code right now.

internal/httpginendpoints/httpginendpoints.go
canfikret: MAA accepts the request with this API version. We can keep it as it is.

PR #253, opened 2026-09-04
Fix Windows ACI token resource encoding

EmanuelOprea:
## Root cause
The Windows ACI token request path uses `url.Values.Encode`, but SKR passes already percent-encoded Key Vault and Managed HSM audiences. This encoded the percent signs a second time, causing the injected identity endpoint to return HTTP 500 during C-WCOW key release.

## Fix
Decode the supplied audience before adding it to the query and add a regression test using the actual Managed HSM value.

## Validation
- `go test ./pkg/common ./pkg/skr`
- Windows `cmd/skr` build
- ACR Windows build from commit `32c7e50`, published at `sha256:4e81b456f52b2d689d3a9b19c5d5abbc9d30c0ad6a947b35967308c1fca8e6d4`
- Live Germany North dashboard run https://github.com/microsoft/confidential-aci-dashboard/actions/runs/34103027206 passed all six C-WCOW SKR tests, including oct-HSM and EC-HSM release, with no Managed Identity token error

### DIFF
--- a/pkg/common/token.go
+++ b/pkg/common/token.go
@@ -77,8 +77,12 @@ func newTokenRequest(resourceId string, identity Identity) (*http.Request, error
 	if err != nil {
 		return nil, errors.Wrapf(err, "parsing managed identity endpoint failed")
 	}
+	resource, err := url.QueryUnescape(resourceId)
+	if err != nil {
+		return nil, errors.Wrapf(err, "decoding managed identity resource failed")
+	}
 	query := uri.Query()
-	query.Set("resource", resourceId)
+	query.Set("resource", resource)
 	if useACIEndpoint {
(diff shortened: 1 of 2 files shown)


micromaomao (approved):
hmm I guess I'm not sure why it works on Linux, maybe the behaviour of go's `query.Set` is different...?

also probably the proper fix is to make these in pkg/skr/skr.go be not URL encoded

```go
	ResourceIdManagedHSM = "https%3A%2F%2Fmanagedhsm.azure.net"
	ResourceIdVault      = "https%3A%2F%2Fvault.azure.net"
```

but I don't really mind either way

micromaomao:
turns out it also wouldn't have worked on Linux and I just didn't test it properly

Thanks!

